PR #114, opened 2019-08-08
a whole brazilian portuguese translation of the presence store descriptions

frozaum:
thanks to @Viruzaum and @Nightdavisao for helping me on this


frozaum:
thanks ❤️ @Fruxh

PR #129, opened 2019-08-10
🔨 Deezer and Twitch Update Metadata

CobyPowers:
Added the new languages to both Deezer and Twitch. Modified the thumbnails and logos for both. Changed tags to remove the name.


CobyPowers:
This is a mess

PR #151, opened 2019-08-14
Add dアニメストア

eai04191:
Japanese not piracy anime streaming service like Crunchyroll.
Probably only available from Japan.

discordapp


eai04191:
Thanks for the translation!

frozaum:
no problem :) @eai04191

PR #153, opened 2019-08-15
added osu!ppy presence

TheSlowly1:


eai04191:
1. seems logo URL in `metadata.json` is wrong.
1. The website contains an official description. You may use it.
    ```
    osu! - Rhythm is just a *click* away!  With Ouendan/EBA, Taiko and original gameplay modes, as well as a fully functional level editor.
    ```
1. largeImageKey always seems to be the same. It would be better to define it when you first create presenceData and then update only details.

PR #177, opened 2019-08-18
German translations for most of the presence description.

OfficialCRUGG:
I hope this doesn't break anything with Bas950's dutch translations, as we basically did them at the same time


frozaum:
oof let me see if i can do something about it

OfficialCRUGG:
Would be great. Maybe you have to add the dutch translations to the conflicting files...

frozaum:
done

PR #218, opened 2019-09-02
Added Showtime Presence

RisingChops:


frozaum:
why you created both js and ts files for the presence lol

PR #278, opened 2019-09-16
ESL update

Bas950:
- [ADDED] www.schulmeisterschaft.de
- [ADDED] Shop now says what you are looking at.
- [BUG - FIXED] Team and Players not showing the right way.
- [BUG - FIXED] News saying ESL Pro instead of ESL News.
- [BUG - FIXED] Forum main page not showing.

All thanks to: Fruxh#3282


Bas950:
@Fruxh 😉

PR #288, opened 2019-09-30
Added WhatsApp.

eggsywashere:
Now using JS paths! It's completely (not 100% sure) safe from leaking non-saved contact numbers to Discord. It can detect if you're typing or can't type and updates your presence as it is.

waiting
typing
canttype
privatenumber


eggsywashere:
Images are from the older version, the title is now "WhatsApp" :3

PR #319, opened 2019-10-05
Update presence.ts

mochiru:


eai04191:
Perhaps this change was made from GitHub's WebUI, but if you change the TypeScript presence, you will also need to update the JavaScript in the dist folder.

CobyPowers:
I'll compile it when I get home, author hasn't updated it in 6 days.

PR #345, opened 2019-10-06
Fixing a bug

callumok2004:


Glitch/dist/metadata.json
```diff
@@ -7,15 +7,15 @@
     "description": {
         "en": "Glitch is the friendly community where you can find or create the app of your dreams. Over a million people have become part of the Glitch community, creating amazing apps, bots, websites, art projects, virtual reality experiences, interactive infographics and much more."
     },
-    "url": ["glitch.com", "support.glitch.com", "status.glitch.com"],
-    "version": "1.2",
+    "url": ["glitch.com", "support.glitch.com", "status.glitch.com", "*.glitch.me", "glitch.me"],
```
Bas950: "*.glitch.me" is not going to work you need to use regex

Glitch/dist/metadata.json
```diff
@@ -7,15 +7,15 @@
...
     "thumbnail": "https://downbeat.is-inside.me/S3bomlzS.png",
     "color": "#f003fc",
     "tags": [
 	"developing",
-	"hosting"
+    "hosting",
+    "community"
     ],
-    "category": "other",
-    "regExp": "([a-z0-9]+)[.]glitch[.]me"
```
Bas950: This was good.

callumok2004: When I used that, anything on status.glitch.com, support.glitch.com  and glitch.com did not work.

TheDropX: It was not working because in the RegExp you put "([a-z0-9]+)[.]glitch[.]me" (".me"), and status.glitch.com, support.glitch.com and glitch.com has ".com" at the endd @callumok2004.

Bas950: Use this "(([a-z0-9]+)[.])?glitch[.]([a-z0-9]+)"
and only have glitch.com in url
that should work for all (both .me and .com and everything in front of glitch

Glitch/dist/metadata.json
CobyPowers: Regexp is what you want here.

frozaum: regexp

Glitch/dist/metadata.json
Bas950: whoops sorry 
(([a-z0-9]+)[.])?glitch[.]([a-z0-9]+)

Glitch/dist/metadata.json
```diff
@@ -18,5 +18,5 @@
     "community"
     ],
     "category": "other",
-    "reqExp": "(([a-z0-9]+)[.])?glitch."
+    "reqExp": "(([a-z0-9]+)[.])?glitch[.]([a-z0-9]+)"
```
Bas950: you typed reQexp instead of reGexp so the right is: regExp

CobyPowers: I have no idea why I didn't catch that.

PR #349, opened 2019-10-06
Added htmlcsscolor presence 🎉

Owen1212055:


Owen1212055:
Note: On line 21 of presence.ts .value always shows as an error. Idk how to fix that. (It works fine tho)

Bas950:
> Note: On line 21 of presence.ts .value always shows as an error. Idk how to fix that. (It works fine tho)

You could fix that by doing 
color = query 
color = color.value

nothing changes just doesnt show the error

PR #373, opened 2019-10-07
fixes start timestamp

Allvaa:


Allvaa:
> Please actually compile the typescript with `tsc` ( or `tsc -w` in watch mode ) or remove it ( and tsconfig ) and work directly with the js file.

I was actually compile the typescript

PR #398, opened 2019-10-08
add more images to resources

Bas950:


Bas950:
writing
call
vcall

Those are the images if you cant see them

Issue #298, opened 2019-10-04
alltheflavors.com


Hexical1:
**What's the name of the service?**
https://alltheflavors.com | AllTheFlavors

**What should the Presence display?**
Current Recipe
Current Flavour
Current Search

**Why do you want support for this service?**
Its widely used by a lot of DIY e-Juice makers and is a helpful website. It would help a lot of others display recipes they are currently browsing and definitely bring more discussions about flavours and recipes in the DIY e juice discord, and others possibly

**Are you able to provide an image for the service? (512x512 minimum)**
https://imgur.com/a/c3Amqof

sunx2:
Added All The Flavors presence .  [Presence Store Link](https://premid.app/store/presences/All%20The%20Flavors)

doomlerd:
@sunx2 thank you.

PR #406, opened 2019-10-08
Added All The Flavors 

sunx2:
its a recipe and DIY site

here are 5 presences but the code can generate presence for every url this site has.

### DIFF
--- a/All The Flavors/dist/metadata.json
+++ b/All The Flavors/dist/metadata.json
@@ -1,0 +1,22 @@
+{
+    "author": {
+        "name": "Nekolen",
+        "id": "270125856679002113"
+    },
+    "service": "alltheflavors",
+    "description": {
+        "en": "alltheflavors.com is a DIY recipe and flavor information site."
+    },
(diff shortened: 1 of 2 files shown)


sunx2:
it also address issue https://github.com/PreMiD/Presences/issues/298

All The Flavors/dist/metadata.json
```diff
@@ -0,0 +1,22 @@
+{
+    "author": {
+        "name": "Nekolen",
+        "id": "270125856679002113"
+    },
+    "service": "alltheflavors",
```
Bas950: Name the service the same as the folder name

sunx2: I did it at first but if i put space in the service, it doesn't load in presence launcher

Bas950: image
https://premid.app/store/presences/Google%20Docs

Works on google docs

sunx2: sure then

PR #407, opened 2019-10-08
Update presence.js

sunx2:
added search system presence
icce cream


sunx2:
changed version from 1.0 to 1.1

PR #410, opened 2019-10-08
created Edge me PLS

sunx2:
as a gift for **fror senpai** .  It has custom dialogue for each instruction the site gives
Screenshot_2
Screenshot_3
Screenshot_4
p1
p3
p4
Screenshot_1


sunx2:
it took so much brainpower to make all possible combinations ;~; . .  all for fror 🤣

PR #430, opened 2019-10-09
:bug: Somebody has to fix this...

sks316:
Prior to this edit, the "Playing back" and "Playback paused" strings for the small image in Astiga were reversed. This pull request fixes that.

Screenshots:
screenie-2019-10-09-18_09_05
screenie-2019-10-09-18_11_03


sks316:
Done lol, can't believe I forgot that

PR #431, opened 2019-10-10
Added Neox Scanlator Presence

Yuujet:
image


Viruzaum:
Neox Scanlator is a Portuguese website, I don't see why it would need to be in English. But it needs a English description

PR #435, opened 2019-10-10
Messenger Console log + User hidden

Bas950:
Same as https://github.com/PreMiD/Presences/pull/423 but that was closed by pullbot


Bas950:
66515756-f6c19880-eadf-11e9-904e-347b1d634ada

PR #445, opened 2019-10-11
Add FnBr.co Presence

NotOfficer:
Simple presence for [fnbr.co](https://fnbr.co)


NotOfficer:
> It's fine for words such as API or the presence name to be in caps though

Ye but looked abit weird when everything is lowercase too.

PR #447, opened 2019-10-11
Fixed a Netflix bug

RisingChops:


RisingChops:
download

RisingChops:
The issue was Black Mirror wasn't showing but this fix fixes the issue

PR #510, opened 2019-10-19
Added p90.zone

frozaum:
First presence o̶r̶ ̶f̶i̶r̶s̶t̶ ̶t̶i̶m̶e̶ ̶i̶ ̶d̶o̶ ̶s̶o̶m̶e̶t̶h̶i̶n̶g̶ ̶i̶n̶ ̶j̶a̶v̶a̶s̶c̶r̶i̶p̶t̶ ̶i̶n̶ ̶m̶o̶n̶t̶h̶s̶  about an epic website that i love to use sometimes

and the screenshots are here:


p90.zone/dist/metadata.json
```diff
@@ -0,0 +1,22 @@
+{
+	"author": {
+		"name": "fror",
+		"id": "515668127829458945"
+	},
+	"service": "p90.zone",
+	"description": "p90.zone is a website that shows random WebMs (mostly memes) uploaded by their users",
```
Bas950: Doesnt this need to be "en:"

frozaum: en is only necessary when it has more than a translation

PR #515, opened 2019-10-20
Added Radio.net

ACertainCoder:
Radio.net

Informations about the current playing radio station

I'm in knowledge of this service for a long time but most people I met didn't know it is existing.

logo_big


frozaum:
Hey @ACertainCoder, to comply with our guidelines we need an example on how the presence shows in your Discord profile, so we can review it.

ACertainCoder:
Screenshot 2019-10-20 195010
Here you go. This is an example of how a radio station is displayed.

PR #519, opened 2019-10-22
Updated dizilab and Game Jolt!

eggsywashere:
Just fixed something in dizilab presence and edited Game Jolt's metadata file. dizilab was having issues when you're browsing through the archive section and I just added Turkish description to the Game Jolt's metadata file. _Only because every presence of mine has Turkish description and this looks awful between others._

Before:

After (yes, big change):
works


eggsywashere:
_Sure, why not._

PR #528, opened 2019-10-24
⭐ Added podcast support & fixed the browsingStamp

ACertainCoder:
Podcasts on Radio.net will now be correctly displayed & fixed the problem with changing song (browsingStamp resetted each time).

Sample on how podcasts will be displayed like:
Discord_2019-10-24_21-11-02


ACertainCoder:
Sorry, I forgot

PR #551, opened 2019-10-30
Trello fix and TruckStopRadio

Bas950:


Bas950:
trello user fix
TSR
tsr2

PR #556, opened 2019-10-31
Updated Presence: Bounce

callumok2004:
Using the new API link, currently, the bounce status does not work on PreMiD, in about an hour I will be coded the backend of the new API link so then it will work! I tested it locally and it works just don't want to expose the API token on PreMiD for security reasons!


callumok2004:
Sr5cBxueV yraAL

callumok2004:
Going to reopen pull request after a few changes have been made

PR #558, opened 2019-10-31
Add SCP Wiki

TheLastZombie:
ye

image
image
image

### DIFF
--- a/SCP Wiki/dist/metadata.json
+++ b/SCP Wiki/dist/metadata.json
@@ -27,13 +27,12 @@
 	"version": "1.0.0",
 	"logo": "https://scp-wiki.wdfiles.com/local--files/component%3Atheme/logo.png",
 	"thumbnail": "https://i.imgur.com/el9c68A.png",
-	"color": "#b01",
+	"color": "#BB0011",
 	"tags": [
 		"fiction",
 		"horror",
 		"reading",
 		"wiki"
 	],
-	"iframe": false,
 	"category": "other"
 }
--- a/SCP Wiki/dist/metadata.json
+++ b/SCP Wiki/dist/metadata.json
@@ -1,0 +1,39 @@
+{
+	"author": {
+		"name": "ry",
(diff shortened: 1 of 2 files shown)


SCP Wiki/dist/metadata.json
```diff
@@ -0,0 +1,39 @@
...
+		"scp-es.com",
+		"scp-th.wikidot.com",
+		"scpclassic.wikidot.com",
+		"scpexplained.wikidot.com",
+		"scp-sandbox-3.wikidot.com"
+	],
+	"version": "1.0.0",
+	"logo": "https://scp-wiki.wdfiles.com/local--files/component%3Atheme/logo.png",
+	"thumbnail": "https://i.imgur.com/el9c68A.png",
+	"color": "#b01",
```
Bas950: Arent color codes 6 long? (hex)

SCP Wiki/dist/metadata.json
```diff
@@ -0,0 +1,39 @@
...
+	"tags": [
+		"fiction",
+		"horror",
+		"reading",
+		"wiki"
+	],
+	"iframe": false,
```
Bas950: might aswell remove the iframe line as its not needed.

SCP Wiki/dist/metadata.json
TheLastZombie: They can be both, see https://www.w3.org/TR/CSS21/syndata.html#color-units:

> The format of an RGB value in hexadecimal notation is a '#' immediately followed by either three or six hexadecimal characters. The three-digit RGB notation (#rgb) is converted into six-digit form (#rrggbb) by replicating digits, not by adding zeros. For example, #fb0 expands to #ffbb00. This ensures that white (#ffffff) can be specified with the short notation (#fff) and removes any dependencies on the color depth of the display.

In this case, `#b01` is equivalent to `#bb0011` and is even being used on the SCP Wiki itself:

image

PR #599, opened 2019-11-08
:star: Support for old version of bilibili

Bas950:


Bas950:
Removes otaku since they shutdown

PR #618, opened 2019-11-12
Added yabancıdizi, another Turkish foreign TV series service!

eggsywashere:
Okay, I don't know what happened but yabancıdizi (actually means foreign TV shows) was the first website I wanted to do if I were going to create a presence in this category. And, when I looked up into it, I figured out that they were using, _iframes in iframes_, I even told this to Timeraa I remember very well. But now, I don't know what changed but seems like it's not like that anymore and runs like a charm. So I made it ^:)

**Homepage:**
homepage

**Viewing a category page:**
looking-at-a-category

**Viewing a user's profile:**
user-profile

**Viewing an actor's page:**
an-actor-page

**Watching an episode:**
watching-a-show

**Watching a movie (yeah, odd but provides movies too):**
watching-a-movie

And here goes fror's thing:
fror-bb

### DIFF
--- a/yabancıdizi/dist/iframe.js
+++ b/yabancıdizi/dist/iframe.js
@@ -1,0 +1,18 @@
+const iframe = new iFrame();
+
+setInterval(function () {
+    const video = document.querySelector("video");
+
+    if (video && video.currentTime && video.duration && video.paused !== undefined) {
+        iframe.send({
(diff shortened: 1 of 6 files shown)


eggsywashere:
Sorry doom :^(

fixed

PR #681, opened 2019-11-24
Gravity presence

jonathanjamandsimi:
Screenshot_6


frozaum:
@doomlerd well i don't thing it is necessary to request a change since many presences uses logos/thumbnails hosted on discord's cdn

PR #716, opened 2019-12-03
Removed WoXAera cuz it doesn't exist anymore.

eggsywashere:
Yeah, how sad is that, check their website: https://woxhub.com


eggsywashere:
Oh, actually, http://woxhub.com now redirects to a website called MemoryHackers, a totally different URL. Delete this anyway, I'm not going to do another one for these guys.

eggsywashere:
And fixed some important stuff for dizilab presence, the issue was:
error

PR #724, opened 2019-12-03
🎉 API down

doomlerd:
thanks @eggsywashere


eggsywashere:
SEE, THIS IS EXACTLY WHY EVERYONE SHOULD USE MY STUFF >:(((
https://eggsy.codes/projects/premid/mdcreator

eggsywashere:
and i am not a presence verifier 😆 blame yourself!

PR #757, opened 2019-12-09
Update (R10.net) & Add Presence (ApertureBot + Türk Hack Team)

atillakurtulussimsek:
**R10.net RPC:** Bugs and Category name fixed!

ApertureBot:
Aperture1
and
Aperture2

Turk Hack Team:
THT

### DIFF
--- a/ApertureBot/dist/presence.js
+++ b/ApertureBot/dist/presence.js
@@ -1,0 +1,58 @@
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
+    function adopt(value) {
+        return value instanceof P ? value : new P(function (resolve) {
+            resolve(value);
+        });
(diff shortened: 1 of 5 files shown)


ApertureBot/dist/presence.js
```diff
@@ -0,0 +1,58 @@
...
+                reject(e);
+            }
+        }
+
+        function step(result) {
+            result.done ? resolve(result.value) : adopt(result.value).then(fulfilled, rejected);
+        }
+        step((generator = generator.apply(thisArg, _arguments || [])).next());
+    });
+};
```
Timeraa: This is not really needed as its created by Typescript, just use the native async await from JS

atillakurtulussimsek: okay!

PR #788, opened 2019-12-14
New icons & Half YT recode

Bas950:
blame @doomlerd


Bas950:
With this half youtube recode we fixed a lot of bugs, but we also found bugs that cant be fixed on our side of things. Youtube caches the channel name, uploader and video title sometimes, so it might show the wrong information once in a while....

PR #811, opened 2019-12-18
Updated clientid

Migushthe2nd:


Migushthe2nd:
Before:
image

After:
image

The diffference is the `:` changed to a `-` in the details row.

PR #859, opened 2020-01-01
Patched Twitch

Alanexei:
Fixed full-screen videos so it no longer displays the "Hosting" message.

Normal Video
Web1
Discord1

Fullscreen Video
Web2
Discord2


Alanexei:
Original report image:
image

PR #875, opened 2020-01-04
Added translation for "Browsing..." in Crunchyroll presence

MrSpinne:
Now making use of the `presence.activity.browsing` key

PR #878, opened 2020-01-05
Added Wakanim presence

MrSpinne:
I wrote a presence to display when watching anime on Wakanim.tv.
Like the Crunchyroll presence, it will display when browsing through the site and when watching an episode.

image
image


Bas950:
You put it in https://github.com/PreMiD/Presences/pull/875 also....

MrSpinne:
> You put it in #875 also....

Yes, thought would be better to separate those 2 presences

MrSpinne:
Closed, I'll make a pull request with both changes again, sorry.

PR #892, opened 2020-01-08
Added Prime Video & Updated Wakanim

### LINKED ISSUE #14: primevideo.com
Ex. https://www.primevideo.com/ which redirects you to amazon.de, amazon.com, amazon.co.jp etc. | Prime Video

Informations about the current series/movie. It would also be great (at least for me), if it would display the title of the series/movie in english.
So instead of for example "Pokémon - Die TV-Serie: Gold und Silber" (which is german) it would display "Pokémon: The Johto Journeys" (which is the english name of it)

Well, Amazon Prime Video, is one of the leading video on demand services, if we ignore Netflix. It would be really cool, if people could see what I am watching on Prime. (I don't really know what else to say here)

http://cdn.crugg.de/files/primevideo-x512.png
http://cdn.crugg.de/files/primevideo-x1024.png
I wasn't able to find an original logo in the right size, so I tried to replicate the logo. I was only able to find a 128x128 favicon by looking through the HTML code of the site. If you can do anything with that, there you are: https://m.media-amazon.com/images/G/01/digital/video/DVUI/favicons/favicon-128x128._CB525481107_.png

MrSpinne:
Prime Video:
- displays when browsing through Prime Video
- displays which video is being watched with progress
- if on amazon.com, the activity will be cleared to avoid conflicts with other presences

image
image

Wakanim:
- Adjusted logo size and color

closes #14


Bas950:
@MrSpinne will it not interfere with the current amazon presence?
There is also https://www.primevideo.com/

You could make it inside of the current Amazon presence I made and adding yourself as a contributor. But it would be nice if you test it to make sure it doesn't interfere with eachother

doomlerd:
I think Prime Video is sort of standalone, after all Amazon is for shopping, different category.

Bas950:
It uses the same domain in some countries tho so I just want to make sure that they will work together @doomlerd

MrSpinne:
https://www.primevideo.com/ has been added.

I've removed Prime Video from the Amazon presence, so it won't interfere.

PR #953, opened 2020-01-24
Switch logo to imgur because pahe's SEO is gae

MulverineX:


animepahe/dist/metadata.json
```diff
@@ -9,7 +9,7 @@
   },
   "url": "animepahe.com",
   "version": "1.0.2",
```
frozaum: you need to change the version number

Alanexei: Increment version number

PR #954, opened 2020-01-24
Add codepen.io

Chezzer-ub:


Alanexei:
Please upload screenshots of the presence and the webpage to verify that it works.

codepen/dist/metadata.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "url": [
+        "codepen.io",
+        "www.codepen.io"
+      ],
+    "description": {
+      "en": "CodePen is an online community for testing and showcasing user-created HTML, CSS and JavaScript code snippets."
+    },
+    "service": "CodePen",
+    "version": "1",
```
Alanexei: This isn't required, but I think you should set this to `1.0.0` instead of `1`. It's more standard.

Chezzer-ub:
image
image
image
image
image

PR #959, opened 2020-01-24
[ Gmail Presence ] Added new features with the author's consent

TellenOfficial:
**_Added new features and changed Gyazo to Imgur. Removed Useless files._**

**I noticed that some functions were missing, so I contacted the author and asked him if I could modify it. I tried to keep his programming style as much as possible. I also changed from Gyazo to Imgur which I think is better. I have attached the image with the author's consent.**

Chat

GMAIL Profile
No Profile GMAIL


doomlerd:
Would be better if you keep the original thumbnail as it's more appealing for the store.

PR #973, opened 2020-01-26
just remove it

Bas950:
We shouldnt allow free sites anymore


doomlerd:
It works fine apparently.

Bas950:
Free domain, the devs changed api for the 3rd time now in 24hours, and the website is as plain as it can be.

PR #984, opened 2020-01-29
✨ Added lain game (laingame.net) and some other stuff

frozaum:
# lain game
lain game is a website trying to reproduce 1998's PlayStation game 'Serial Experiments Lain' in web form.
This game has no gameplay.
The player must check each file one by one from bottom levels to the upper and assemble storylines from pieces of information.

# screenshots
index/main page
about
viewing a file

# Dora the Explorer
* rip dora

# Domino's Pizza (UK)
* just changing the name of the presence


frozaum:
^^ problem i have with this is that randomly it choses flash or html, so maybe it will happen.

PR #1028, opened 2020-02-09
⭐ Added IFTTT

ACertainCoder:
IFTTT

Current information on what you are creating, editing, viewing, ...

It helps you to let your apps and devices work together for free.

logo_big

**What should the presence look like when active?**
Discord_1581237183
Discord_1581237323


ACertainCoder:
Please ignore the fact that I've already pulled another presence and removed it again. I just realised too late that it already existed.

PR #1032, opened 2020-02-10
mangadex

Shironep:
fixed chapter info


Shironep:
Chapter with title
VolCh with title
Ch without title

PR #1058, opened 2020-02-14
📝 [OGLLE.COM.BR] Presence Version 1.1

MateusOFCZ:
> New logo and thumbnail;
> Tidy English description;
> Problems showing some pages in the tidy presence.

### DIFF
--- a/Oglle/dist/metadata.json
+++ b/Oglle/dist/metadata.json
@@ -6,12 +6,12 @@
 	"service": "Oglle",
 	"description": {
         "pt_BR": "Precisa de um BOT para Discord? Edição/criação de sites? Ou qualquer outro serviço fora do comum? Este é o site feito para você!",
-        "en": "Need a BOT for Discord? Website editing/creation? Or any other out-of-the-ordinary service? This is the site made for you!"
+        "en": "Need a BOT for Discord? Website editing/creation? Or any other out of the ordinary service? This is the site made for you!"
 	},
-	"url": "oglle.com.br",
-	"version": "1.0",
-	"logo": "https://i.imgur.com/OgCAj9v.png",
-	"thumbnail": "https://i.imgur.com/Cr52ZSV.png",
+	"url": ["oglle.com.br"],
+	"version": "1.1",
+	"logo": "https://i.imgur.com/MXfV3G8.png",
+	"thumbnail": "https://i.imgur.com/tPXy6IS.png",
 	"color": "#005eff",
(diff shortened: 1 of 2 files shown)


Oglle/dist/metadata.json
```diff
@@ -6,13 +6,13 @@
...
+        "en": "Need a BOT for Discord? Website editing/creation? Or any other out of the ordinary service? This is the site made for you!"
 	},
-	"url": "oglle.com.br",
-	"version": "1.0",
-	"logo": "https://i.imgur.com/OgCAj9v.png",
-	"thumbnail": "https://i.imgur.com/Cr52ZSV.png",
+	"url": ["oglle.com.br"],
+	"version": "1.1",
+	"logo": "https://i.imgur.com/MXfV3G8.png",
+	"thumbnail": "https://i.imgur.com/tPXy6IS.png",
```
Alanexei: Keep the other thumbnail, this one is too small.

Oglle/dist/presence.js
```diff
@@ -10,7 +10,10 @@ presence.on("UpdateData", () => {
     };
     presenceData.startTimestamp = browsingStamp;
 
-    if(window.location.pathname.includes("")) {
+    if(window.location.pathname.includes()) {
```
Alanexei: Why are you doing this?

MateusOFCZ: The start page is not showing up in the presence, I added this line and it started to work as expected!

Oglle/dist/presence.js
```diff
@@ -10,7 +10,10 @@ presence.on("UpdateData", () => {
     };
     presenceData.startTimestamp = browsingStamp;
 
-    if(window.location.pathname.includes("")) {
+    if(window.location.pathname.includes()) {
+        presenceData.details = "Inicio"
+    }
```
Bas950: Can just be removed since it doesnt do anything.

MateusOFCZ: If removed, the home page will not be shown in the user's status with our presence, since the home page does not need to have "/home" or in this case in Portuguese "/inicio"!

Bas950: image
This one should fix that tho.. Try it.

MateusOFCZ: I already tested it, but sometimes there is no "/" in the url, then it does not recognize it as the beginning, in this case if it does not have the "/" it will be the beginning, but if it does, it will includes("/") when the url has "/" it will not appear as a start, and if it leaves only includes() when it has "/" it will not appear as a start either...

PR #1064, opened 2020-02-18
Roblox Developer Forum | devforum.roblox.com

Swardoric:
**Prerequisites and essential questions** 
* [x] Is it a popular site?
* [x] Is the website older than 2 months? 
* [ ] Is the site locked to a specific country/region?
* [ ] Is the site a paid service? (e.g. Netflix, Hulu)
* [ ] Does the website feature NSFW content? (e.g. porn, etc...)
* [ ] Are you a donator/patron?
* [ ] Do you acknowledge that coding presences is completely voluntary and may take time for your service to be added regardless of priority?

**What's your Discord username?** 
Swardoric#9725
Roblox Developer Forum
Presence will display the post you are reading, the page you are on, the category you are in and if you are set up your account.
**If possible, please provide a logo for the service (512x512 minimum)** 
RDF_icon

### DIFF
--- a/Roblox Developer Forum/dist/metadata.json
+++ b/Roblox Developer Forum/dist/metadata.json
@@ -1,0 +1,18 @@
+{
+    "author": {
+      "name": "Swardoric",
+      "id":371652403071811595
+    },
+    "service": "Roblox Developer Forum",
+    "description": {
+      "en": "A place where Roblox's developer community can share their thoughts about creating games and get updates from staff members.",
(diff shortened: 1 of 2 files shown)


Bas950:
Why dont you just update the current roblox presence and add this?

Swardoric:
> Why dont you just update the current roblox presence and add this?

The Forum is different from the site, that's why I create a presence for it

Alanexei:
Some websites have forums, byte.co for example. You should have still added it to the roblox presence. Also you should remove the heavy lifting function.

doomlerd:
The forum is basically just a sub-domain, and they're supported too but not when they can be part of the main domain. Sorry, but you'll have to revert these changes and add it to the main presence ( don't forget to increment the version! )

PR #1099, opened 2020-02-26
Gitea Presence

OfficialCRUGG:
Added a Presence for the Gitea Website.

image
image

P.S. I didn't do a presence in a long time. While the code should work, I may have done something wrong in the metadata, so please check if you can find any errors.

### DIFF
--- a/Gitea/dist/metadata.json
+++ b/Gitea/dist/metadata.json
@@ -8,9 +8,7 @@
         "en": "Gitea is an Open-Source Git Platform heavily inspired by GitHub",
         "de": "Gitea ist eine Open-Source Git Plattform, welche sehr stark an GitHub angelehnt ist"
     },
-    "url": [
-        "gitea.com"
-    ],
+    "url": "gitea.com",
     "version": "1.0",
     "logo": "https://crugg.de/premid/gitealogo.png",
     "thumbnail": "https://crugg.de/premid/giteabanner.png",
--- a/Gitea/dist/metadata.json
+++ b/Gitea/dist/metadata.json
@@ -1,0 +1,26 @@
+{
+    "author": {
+        "name": "CRUGG",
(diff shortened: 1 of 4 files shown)


Gitea/dist/metadata.json
```diff
@@ -0,0 +1,26 @@
...
+        "id": "228965621478588416"
+    },
+    "service": "Gitea",
+    "description": {
+        "en": "Gitea is an Open-Source Git Platform heavily inspired by GitHub",
+        "de": "Gitea ist eine Open-Source Git Plattform, welche sehr stark an GitHub angelehnt ist"
+    },
+    "url": [
+        "gitea.com"
+    ],
```
Bas950: if there is only 1 url it shouldn't be in an array

OfficialCRUGG: I did that because in the very near future I also want to support their homepage gitea.io. I can also change it for now though.

OfficialCRUGG:
@doomlerd Why do I have to use imgur?

Bas950:
> Use imgur for the logo and thumbnail.

Thats not needed? He uses a custom one, its not from the site itself.

Gitea/dist/metadata.json
Bas950: For now you should change it.

doomlerd:
> @doomlerd Why do I have to use imgur?

It's more reliable for the store than a custom image host at the moment.

OfficialCRUGG:
All my other presences use my own host as well and I would like to keep it that way, since there is no guideline to use imgur. Also imgur has been down a lot recently so in the last few months my host had a way better uptime.

doomlerd:
> I would like to keep it that way, since there is no guideline to use imgur.
Makes sense

PR #1101, opened 2020-02-27
taiga Bot Presence

OfficialCRUGG:
PreMiD Presence for the taiga Discord Bot.
This presence uses multiple Applications for different Displaynames in Presence, however all are used for Websites owned by the taiga Bot or of Sub-Projects which are made by taiga.
This has been tested a lot and has already been used in another Presence of mine before.

image
image
image
image
image

### DIFF
--- a/taiga Bot/dist/metadata.json
+++ b/taiga Bot/dist/metadata.json
@@ -29,8 +29,7 @@
     "color": "#ff006d",
     "tags": [
         "discord",
-        "bot",
-        "trading"
+        "bot"
     ],
     "category": "other"
 }
--- a/taiga Bot/dist/metadata.json
+++ b/taiga Bot/dist/metadata.json
@@ -1,0 +1,36 @@
+{
+    "author": {
+        "name": "CRUGG",
+        "id": "228965621478588416"
+    },
+    "service": "taiga Bot",
+    "description": {
+        "en": "taiga is a Discord Bot with many features like Virtual Trading Cards, an economy system and in the future even a Plugin Marketplace",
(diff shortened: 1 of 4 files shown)


taiga Bot/dist/metadata.json
```diff
@@ -0,0 +1,35 @@
...
+        "git.taigabot.net",
+        "labs.taigabot.net",
+        "legal.taigabot.net",
+        "taipl.taigabot.net",
+        "telegram.taigabot.net",
+        "translate.taigabot.net"
+    ],
+    "version": "1.0",
+    "logo": "https://cdn.taigabot.net/assets/branding/taiga/premid_logo.png",
+    "thumbnail": "https://cdn.taigabot.net/assets/branding/taiga/premid_banner.png",
```
Bas950: upload to imgur.

OfficialCRUGG: There is no presence guideline which says that "its not allowed to be on the cdn of the site" like you told me on Discord. Also there is like literally no difference between me hosting images on my private site and me hosting images on my project's site. They both run on the same VPS and even use the same Webserver Instance.

Alanexei: Imgur retains images forever, disregarding the hiccups here and there. But there's no guarantee that the CDN will keep the name of the logo the same. For example, it could be changed to `icon.png` on the cdn and will break the presence's icon.

Alanexei: You did use your private CDN on another presence so it'll be fine to use this one.

OfficialCRUGG: @Alanexei The only way the icon name could change would be that I use an SFTP Client, login via SSH or use our Admin Interface and manually change it.

PR #1114, opened 2020-02-29
⭐ Guideline Modifications

Alanexei:


doomlerd:
better clarify the part where rewriting isn't allowed, like contributors should be allowed to provide hotfixes. just not try to fix something when it's not broken.

CONTRIBUTING.md
```diff
@@ -75,5 +75,6 @@ For your presence to reach the stores, it MUST go through a process on GitHub to
 11. The `url` field **MUST NOT** include `http://` or `https://`, neither the parameters (e.g. a presence for `https://www.google.com/search?gws_rd=ssl` will only have `www.google.com` in the `url` field).
 12. Descriptions and tags should always be in arrays, even when it's only one element. The `url` field, however, should only be a string if it's one domain.
 13. Unstable sites that constantly change APIs/domains, randomize HTML elements or just still being in heavy development are not allowed and will be removed from the store.
+14. The logo and thumbnail on your presence **MUST** either be `i.imgur.com` or from a CDN that **YOU** own. You are allowed to use the CDN of the website **ONLY** if you are an owner or an administrator and have the proper permissions to make changes to that image.
```
frozaum: what about other image hosting services like [kafuka.moe](https://kafuka.moe), [owo domains](https://whats-th.is/faq.html), [is-inside.me](https://is-inside.me/) and many, many others.

doomlerd: "cdn that you own" is part of that or i'm wrong?

Alanexei: I'll keep `is-inside.me`, but there are many hosting options, but figuring out which ones are valid may cause headaches for the verifiers slowing down verification times. Presences that use these hosts can keep them, just new ones will have to follow the new set of guidelines.

Alanexei: If you have "access" to deleting and changing files from those domains then yeah it's fine. I'll mention that.

PR #1120, opened 2020-03-01
Update metadata.json

iammanish17:


doomlerd:
Don't forget viaplay.se and viaplay.dk

PR #1146, opened 2020-03-06
Plex: Add season and chapter indicator

Rubensei:
Chapter name isn't informative enough when you are watching a tv show
tv-show
Checked to avoid displaying unwanted/duplicated info on other types of media
movie
 PS: Also tested music, news, podcasts


Bas950:
Why 04x22 instead of S4E22?

Bas950:
And you can add yourself as contributor if you want.

Rubensei:
> Why 04x22 instead of S4E22?

I'm used to that format tbh, never liked the plex one that much but don't really care about the format, just about the info

Bas950:
> I'm used to that format tbh, never liked the plex one that much but don't really care about the format, just about the info

S and E are more used and easier to understand

Rubensei:
I'm changing it right now, also editing the .ts instead of the .js directly

Bas950:
Why did you remove all ; ?

Rubensei:
I think it was my damn linter 🤦‍♂ 
Although they are not necessary in typescript (only in a few situations and my linter didin't notice any of them in the code 🤔 )

Alanexei:
I think the way it was before was fine. The change of the format is a general modification rather than a bug fix.

PR #1151, opened 2020-03-07
Added DaFont support

Jakeisbored:
Examples :
image
image
image
image

### DIFF
--- a/DaFont/dist/metadata.json
+++ b/DaFont/dist/metadata.json
@@ -1,0 +1,21 @@
+{
+    "author": {
+        "name": "Jake.#1111",
+        "id": "480407581085532180"
+    },
+    "service": "DaFont",
+    "description": {
+        "en": "Archive of freely downloadable fonts. Browse by alphabetical listing, by style, by author or by popularity."
+    },
+    "url": "www.dafont.com",
+    "version": "1.0",
+    "logo": "https://i.imgur.com/rc9ODvW.png",
+    "thumbnail": "https://i.imgur.com/OTjFjMj.jpg",
+    "color": "#eb4034",
+    "tags": [
+        "fonts",
+        "design",
+        "text"
+    ],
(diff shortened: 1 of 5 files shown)


DaFont/dist/metadata.json
```diff
@@ -0,0 +1,21 @@
+{
+    "author": {
+        "name": "Jake.#1111",
```
eggsywashere: Please remove your user tag (#*1111*) from here.

Jakeisbored: > I never got that before  . Is this new?

Bas950: Nope always been there, if it is in your others, please change them. We might've missed it.

Jakeisbored: Yeah you did, will do. For now,  I've fixed this one

eggsywashere: If you could also fix the others and increment their versions, you can commit them to your fork and this PR will merge all of them. There won't be need for another PR. It's up to you.

DaFont/dist/presence.js
```diff
@@ -0,0 +1,74 @@
...
+    mediaKeys: false
+});
+
+presence.on('UpdateData', () => {
+    const presenceData = {
+            largeImageKey: "dafont"
+        },
+        Path = document.location.pathname,
+        Details = {};
+    presenceData.startTimestamp = Math.floor(Date.now() / 1000)
```
eggsywashere: This is optional but for the "quality" of your code, you may want to move the `startTimestamp` directly into your `presenceData` object since you're not changing it anywhere else, there's no reason to set this outside of it. It's okay if you want to keep it but it'll save a line!

frozaum: ^^

Jakeisbored: Sure!

PR #1158, opened 2020-03-10
Basic Cleanup of the GBAtemp presence

NightYoshi370:
Needs testing, but it should be good to go. I'll look over it later to see if I could clean anything else up

### DIFF
--- a/GBAtemp.net/presence.ts
+++ b/GBAtemp.net/presence.ts
@@ -1,65 +1,53 @@
-let timeElapsed = Math.floor(Date.now() / 1000),
-  threadName,
-  authorName,
-  newsAuthor,
-  profileName,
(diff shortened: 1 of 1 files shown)


NightYoshi370:
image

NightYoshi370:
To what? To the end user, this is exactly the same but the entire backend has been rewritten.

Alanexei:
Change the version in the metadata. It's the way the server updates presences that requires the version to be changed.

Alanexei:
Another thing, the only file changed was the typescript. Make sure you compile the typescript file and commit the changes.

NightYoshi370:
I'm too busy unfortunately to do all this. I'll close the PR myself, and re open when I could do it again

PR #1163, opened 2020-03-12
Add TVNZ OnDemand Presence

Cairo2k18:
Examples:

### DIFF
--- a/TVNZ OnDemand/dist/metadata.json
+++ b/TVNZ OnDemand/dist/metadata.json
@@ -1,0 +1,24 @@
+{
+    "author": {
+      "name": "Cairo",
+      "id": "208105877838888960"
+    },
+    "url": [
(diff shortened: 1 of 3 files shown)


TVNZ OnDemand/dist/metadata.json
```diff
@@ -0,0 +1,24 @@
+{
+    "author": {
+      "name": "Cairo",
+      "id": "208105877838888960"
+    },
+    "url": [
+      "www.tvnz.co.nz"
+    ],
```
Bas950: url shouldnt be in an array if it is only 1

Cairo2k18: > url shouldnt be in an array if it is only 1

Thanks for pointing that out, it originally contained a non-www url as well but I removed it and forgot to remove the array.

PR #1173, opened 2020-03-13
Youtube presence fix & upgrade

Rubensei:
Fixed the issue with duplicated channel name while viewing a video
Before fix:
image

After fix:
image

Added support for community post view page
image

EDIT: Added fix images


doomlerd:
that's afaik only the TV version you updated which really isn't broken right now

Rubensei:
Although that's actually fixing the issue, because if you check the lines 114-119 on `presence.ts`, the state assignment uses that value before any other

Rubensei:
Sorry totally forgot the version 🤦‍♂

Bas950:
uhm also upload a screen of the normal video thing working

PR #1192, opened 2020-03-20
Disney+: update tags, thumbnail, description in more langs

ThaTiemsz:


ThaTiemsz:
Forgot to change that, my bad.

PR #1196, opened 2020-03-20
First version of Yarn presence

EvilBrain:
image


Yarn/dist/metadata.json
```diff
@@ -0,0 +1,25 @@
+{
+    "author": {
+      "name": "Deleted 7.12i",
+      "id": "591702135310385157"
+    },
+    "url": [
+      "yarnpkg.com"
```
Alanexei: Make the url a string not an array

PR #1201, opened 2020-03-21
Radiogarden new website layout fix

tepiloxtl:
Fixed presence to work with new website layout. Added statuses for not playing, playing and tuning
rg11


tepiloxtl:
Oh, actually I have an idea, but I need to test it out in the morning, so closing for now...

PR #1226, opened 2020-03-24
Revert "rip bounce"

callumok2004:
Reverts PreMiD/Presences#1096

### DIFF
--- a/Bounce/dist/metadata.json
+++ b/Bounce/dist/metadata.json
@@ -1,14 +1,8 @@
 {
   "author": {
-    "name": "Bas950",
-    "id": "241278257335500811"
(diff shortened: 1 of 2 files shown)


callumok2004:
Already incremented version -.-
And Bas, I am literally a developer for bounce haha, the site is not **fully** released, but the API is up for the stats to work, which you can see if you check the API link in the presence itself -.-

callumok2004:
image

doomlerd:
Why did you change the API link?

callumok2004:
Were changing up our API behind the scenes soon, with that new api link, and changes we make will go over that one aswell so it dont break the presence

PR #1228, opened 2020-03-25
Update TruckersFM Presence

pascalla:
- Updated Artwork
- Updated Description
- Updated Colour
- Updated ags

### DIFF
--- a/TruckersFM/dist/metadata.json
+++ b/TruckersFM/dist/metadata.json
@@ -5,13 +5,13 @@
   },
   "service": "TruckersFM",
   "description": {
-    "en": "TruckersFM is the  #1 community station for Trucking Simulators. TruckersFM provides a variety of shows from live presenters 24/7, meaning no matter where you are from, the tunes will keep on playing!"
+    "en": "Truckers.FM is the  No.1 Hit Music Station as voted by the TruckersMP Community 5 years running. Broadcasting live online, on TuneIn and on your smart speaker, Truckers.FM provides round the clock entertainment broadcasting a variety of shows, ranging from the Top 40 to the biggest throwbacks. Be sure to follow us on our social media channels @TruckersFM"
   },
   "url": "https://truckers.fm",
   "version": "1.0.3",
-  "logo": "https://ninjalabs.co.uk/i/p0ezmnp.png",
-  "thumbnail": "https://ninjalabs.co.uk/i/i48db0v.png",
-  "color": "#fc0303",
+  "logo": "https://i.imgur.com/f4JkWUI.png",
+  "thumbnail": "https://i.imgur.com/vMj6Rcz.png",
+  "color": "#E1306C",
   "tags": [
     "TFM",
     "TruckersMP".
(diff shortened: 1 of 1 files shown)


TruckersFM/dist/metadata.json
```diff
@@ -1,23 +1,25 @@
 {
   "author": {
-    "name": "Ninja",
-    "id": "150987570715688960"
+    "name": "Pascal",
+    "id": "156560379089125376"
```
Bas950: You shouldn't change the author, you can add yourself as a contributor

TruckersFM/dist/metadata.json
```diff
@@ -1,23 +1,25 @@
...
-    "en": "The number one hit music station for Trucking Simulators. Radio made for driving, now playing online, in cab and on the go."
+    "en": "Truckers.FM is the  No.1 Hit Music Station as voted by the TruckersMP Community 5 years running. Broadcasting live online, on TuneIn and on your smart speaker, Truckers.FM provides round the clock entertainment broadcasting a variety of shows, ranging from the Top 40 to the biggest throwbacks. Be sure to follow us on our social media channels @TruckersFM"
   },
-  "url": "truckers.fm",
-  "version": "1.0.2",
-  "logo": "https://ninjalabs.co.uk/i/p0ezmnp.png",
-  "thumbnail": "https://ninjalabs.co.uk/i/i48db0v.png",
-  "color": "#fc0303",
+  "url": "https://truckers.fm",
+  "version": "1.0.3",
```
Bas950: url shouldnt include https://

TruckersFM/dist/metadata.json
```diff
@@ -1,23 +1,25 @@
...
+  "color": "#E1306C",
   "tags": [
     "TFM",
+    "TruckersMP",
     "radio",
+    "music",
     "ETS2",
     "ATS",
-    "Your Drive, Your Music"
+    "Simulator Radio"
```
Bas950: tags are not allowed to have spaces anymore use -

pascalla:
> Also this only updates the metadata is that correct?

Yeah, we noticed the images were hosted on a self upload site by original creator which seems to have invalid certs, also took opportunity to have some nicer images :)

Bas950:
> Yeah, we noticed the images were hosted on a self upload site by original creator which seems to have invalid certs, also took opportunity to have some nicer images :)

All good, regarding the other tags, the rule changed a week ago so I looked it up:
• Tags should be used as alternate names whenever possible, shortened versions must be included as well (e.g. if an Amazon presence had included AWS support it would have its tags like : "amazon-web-services" and "aws"). This is not required if it's not possible, but will make it easier for users when searching.
• Tags must not include any spaces, slashes, single/double quotation marks, unicode characters and should always be lowercase.

Could you update the other tags as well?

pascalla:
> Make the description shorter ( x3 ) and remove the part where you promote your social media account.

Hope changes are satisfactory

PR #1229, opened 2020-03-25
⭐ Updated Radio.net

ACertainCoder:
In version 2.2 of Radio.net, I've added missing pages and improved some stuff:

- Removed the "language"-section (I recoded the stuff that previously needed it)

- New thumbnail
thumbnail

- Better station informations while the player is paused

- Displays if the user is watching an ad

- Better representation of the search

- Added "My Favorites", "Recently Played", "My Profile" and the smartphone apps


Radio.net/dist/metadata.json
```diff
@@ -1,38 +1,38 @@
...
+    "www.radio.dk",
+    "www.radio.pt",
+    "www.radio.es",
+    "www.radio.it",
+    "www.radio.pl",
+    "br.radio.net",
+    "mx.radio.net",
+    "co.radio.net"
+  ],
+  "version": "2.2",
```
Bas950: maybe use regex? 
And you need to update the version

ACertainCoder: Oh, yes no problem. I haven't seen that there was a version bump.

PR #1235, opened 2020-03-26
Disney+: use regex for url matching

ThaTiemsz:
This should hopefully fix the presence not working for some people.


Disney+/dist/metadata.json
```diff
@@ -12,8 +12,8 @@
     "it": "Scopri il meglio di Disney, Pixar, Marvel, Star Wars e National Geographic. Film, serie, produzioni originali e molto altro.",
     "es": "El servicio de streaming de Disney, Pixar, Marvel, Star Wars y National Geographic."
   },
-  "url": "www.disneyplus.com",
-  "version": "1.2.3",
+  "regExp": "(?:www[.])?disneyplus[.]com",
+  "version": "1.2.4",
```
Bas950: you must also keep the url, as its used as a backup if for some reason the regexp fails

PR #1258, opened 2020-03-30
Update presences repo ( rebased )

doomlerd:
A try to ignore deepscan's false positives


doomlerd:
# Current progress & changes

### FIXED :  
[check the previous commits, but technically 90% 🎉]

### DELETED : 
`Neox Scanlator` - full of JS errors | definitely needs to be recoded
`Newgrounds` - iframe not working | can be fixed without recoding

`Super Animes` - +60 errors within one file | needs to be PROPERLY recoded
`Super Hentais` - same as above
`Super Mangas` - same as the two above

`Tidal` - the whole presence checks for strings in booleans | can be fixed without recoding
`Codeforces` - broken | can be fixed without recoding
`DeviantArt` - does not respect TS standards | better coded in native JS

`HentaiHeaven` - broken | can be fixed without recoding

`animeflix` - compiled without source files | can be fixed without recoding, decision goes to the author
`IFTTT` - same as above
`Keep Talking and Nobody Explodes` - same as above
`My Nintendo` - same as above
`p90.zone` - same as above
`Radio.net` - same as above
`RadioMe` - same as above
`Rythm` - same as above
`s.to` - same as above
`Vlare` - same as above

`DockerHub` - Duplicate declarations across the whole presence | can be fixed without recoding
`WIMS` - same as above

doomlerd:
🎉

PR #1257, opened 2020-03-30
Update DeviantArt presence (1.1.0, RC 1)

Hans5958:
## Preword

Hello, it's been a while since the last update of this presence, and I finally did it. 

This update introduces 10 supported domains outside www.deviantart.com, which I don't know if a lot of people use it, except sta.sh. I think there are more than 6k lines added within this update, along with refactors here and there.

This updates have been done since weeks ago, but I finally pushed it because of the changes requested. So, here you go.

## Changelog

- Changed things for consistency
- Supported more domains
	- about.deviantart.com
	- chat.deviantart.com
	- forum.deviantart.com
	- groups.deviantart.com
	- portfolio.deviantart.com
	- shop.deviantart.com
	- www.deviantartsupport.com
	- eclipsefeedback.com
	- deviantartads.com
	- sta.sh
- Supported more pages on www.deviantart.com
- Added multiple code functions
- Changed default presence texts (request from staff members)

## Notes

- In my side, sometimes the presence didn't work and requires reloading. I think this happened because I used Edge and my laptop is kind of acting weird. It should work 100% in theory.
- I removed the ``null``s into ``undefined``. Hope you guys like it.
- I'm going to request this one first, and then do the other requests, in case you need more changes.

## Images

| Image | Link visited |
| -- | -- |
||https://www.deviantart.com/
||https://www.deviantart.com/chychydapotats/art/Redraw-bEEt-831932372 (sponsored link)
|https://about.deviantart.com/|
||https://chat.deviantart.com/
||https://forum.deviantart.com/
||https://forum.deviantart.com/devart/general/2575523/
||https://groups.deviantart.com/
||https://portfolio.deviantart.com/
||https://shop.deviantart.com/
||https://www.deviantartsupport.com/en/
||https://www.deviantartsupport.com/en/about-deviantart-2821724/about-us
||https://www.eclipsefeedback.com/
||https://deviantartads.com/
||https://sta.sh/
||https://sta.sh/01u0obvy1hn6

PR #1267, opened 2020-03-31
Update DeviantArt presence (1.1.0, RC 2)

Hans5958:
**(reviewers: please read notes)**

## Preword

Hello, it's been a while since the last update of this presence, and I finally did it. 

This update introduces 10 supported domains outside www.deviantart.com, which I don't know if a lot of people use it, except sta.sh. I think there are more than 6k lines added within this update, along with refactors here and there.

This updates have been done since weeks ago, but I finally pushed it because of the changes requested. So, here you go.

### Release Candidate 1

- Changed things for consistency
- Supported more domains
	- about.deviantart.com
	- chat.deviantart.com
	- forum.deviantart.com
	- groups.deviantart.com
	- portfolio.deviantart.com
	- shop.deviantart.com
	- www.deviantartsupport.com
	- eclipsefeedback.com
	- deviantartads.com
	- sta.sh
- Supported more pages on www.deviantart.com
- Added multiple code functions
- Changed default presence texts (request from staff members)

### Release Candidate 2

- Changed variable scopes to fix DUPLICATE-DECL on DeepScan. (``var`` to ``let``)

## Notes

- In my side, sometimes the presence didn't work and requires reloading. I think this happened because I used Edge and my laptop is kind of acting weird. It should work 100% in theory.
- I removed the ``null``s into ``undefined``. Hope you guys like it.
- I'm going to request this one first, and then do the other requests, in case you need more changes.
- This update includes all DeepScan fixes from #1257. (it's a low priority guys!!)
- **I still don't understand that this presence is "not respect TS standards." Even after running the compile command, nothing new appears, just this one error about TS didn't know the browser or something. So, please help.**

## Images

| Image | Link visited |
| -- | -- |
||https://www.deviantart.com/
||https://www.deviantart.com/chychydapotats/art/Redraw-bEEt-831932372 (sponsored link)
|https://about.deviantart.com/|
||https://chat.deviantart.com/
||https://forum.deviantart.com/
||https://forum.deviantart.com/devart/general/2575523/
||https://groups.deviantart.com/
||https://portfolio.deviantart.com/
||https://shop.deviantart.com/
||https://www.deviantartsupport.com/en/
||https://www.deviantartsupport.com/en/about-deviantart-2821724/about-us
||https://www.eclipsefeedback.com/
||https://deviantartads.com/
||https://sta.sh/
||https://sta.sh/01u0obvy1hn6


Hans5958:
Can I have a follow-up?

DeviantArt/presence.ts
```diff
@@ -0,0 +1,619 @@
+var presence = new Presence({
+	clientId: "664057766809436161",
+	mediaKeys: false
+})
```
Bas950, line 4: remove media keys

Hans5958: > 
> 
> remove media keys

Will do, thanks.

Hans5958:
> remove the custom functions, native variants exist ( make sure to check the docs/presence class )


> 
> 
> remove the custom functions, native variants exist ( make sure to check the docs/presence class )

Could you give me an example? I read the docs but I'm not sure what you're referring.

doomlerd:
> Could you give me an example? I read the docs but I'm not sure what you're referring.

The clearData function, also about your notes, undefined is technically the same as null.

Hans5958:
> > Could you give me an example? I read the docs but I'm not sure what you're referring.
> 
> The clearData function, also about your notes, undefined is technically the same as null
> and just a friendly reminder, Fandom is deleted as well, you may include it here and apply the same fixes.

The ``clearData`` function is already removed on this pulley. I think the only difference about ``null`` and ``undefined`` is that ``null`` threw errors and ``undefined`` isn't. So, I don't need the ``clearData`` function at all, and I just wanted to put that thing as a quick reference for me, if you don't mind.

I am aware about Fandom and I have done changes since two days ago. I'll just wait until this pulley got accepted just in case you guys need to do some changes here that also applies to any of my presences.

Hans5958:
what
firefox_2020-04-04_14-52-45

Hans5958:
> Still, you have to remove the undefined strings, whether it throws errors or not, find a better way.
> and update your fork to get rid of the errors, closing this PR in 2 days.
The ``undefined`` values won't outputting errors, as I said, and that's the best and only way to have that variable **without** additional function (``clearData``) and having a value.

You know what? Fine, I'll remove it, it won't make any difference when it's run nonetheless.

> _and no, that function is still there, it's resetData_

Okay, I don't want to write another long explanation, but I supposed I need to explain this because of the novelty that I made and this misundestanding.

**``presence.clearActivity()`` is different than ``resetData()``**.

1. ``presence.clearActivity()`` makes PreMiD stops the Rich Presence, removing the presence activity at all.
2. ``resetData()`` resets the variable ``presenceData``.

Now, I only need to **reset** the value of ``presenceData,`` **not** removing the whole presence activity from Discord.

I suggest you to read this explanation so you really understand what's going on.
https://github.com/Hans5958/PreMiD-Presences-Personal/blob/master/%23%20Base/explanation-for-base.md

doomlerd:
I understand that you only need to send the changed data and not everything at once, but that's something that can be implemented on the extension level. Your Presence is really broken, can't just say that the issues every single person who made a ticket about one of your Presences faced are on their end.
It may seem to you that your code is clean, but it isn't ( and was deleted for a reason, which i've already explained before ), that function is definitely the same as `clearActivity()` as it's used for the same purpose, it simply runs locally while the other one runs right before being sent to Discord; and even then, it didn't work for the users I've mentioned earlier.
I'll let this slip through, thanks to @Alanexei. But I can't guarantee that your Presence ( or any similar to it ) will stay on the store if we receive more tickets about the same issue, since it doesn't comply with the guidelines.

PR #1280, opened 2020-04-03
Fixed the issues with the WIMS presence

Migushthe2nd:
The presence still is an absolute mess, but it works. I might update it in the coming weeks/months.
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "deepscan.enable": true
+}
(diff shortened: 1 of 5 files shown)


Migushthe2nd:
Oh yeah my fault, didn't mean to include the vscode settings file. Yes you are right, I shouldn't have increased the version and the presence only works on the default theme, because WIMS' code is absolute chaos. A WIMS server can be configured to use different themes, but that also changes the complete layout of the website, classes and ids. Hopefully it's all good now.

PR #1282, opened 2020-04-03
Fix Custom Status

eggsywashere:


eggsywashere:
omg it's owrking ! ! @doomlerd @Bas950 

image

eggsywashere:
image
hapi

doomlerd:
that first screenshot really seems like it's from the website not the presence itself xD

eggsywashere:
> that first screenshot really seems like it's from the website not the presence itself xD

not going to reply to that..

PR #1291, opened 2020-04-05
Update metadata.json

ModBotStudio:
added ` "button": false`


Iloot.it/dist/metadata.json
```diff
@@ -10,13 +10,14 @@
   "url": [
     "iloot.it"
   ],
```
Bas950, line 12: url shouldnt be in array if only 1

ModBotStudio: oh wait i forgot

PR #1328, opened 2020-04-14
Update DeviantArt presence (1.1.1)

Hans5958:
## Preword

Here's the recode for DeviantArt so it is valid in the eyes of ESLint, as per request of @Bas950.

For others, due to the length of the code for this presence, ESLint was pretty mad and spitting a ton of errors and warnings. I agree to fix it and remove the presence for the store momentarily (b5209db0). I have since resolved the ESLint problems, along with other presence that are in queue on making its PR, just for consistency.

## Changelog

- Refactored code for ESLint validation

## Notes

I forgot to reply to @doomlerd about the `resetData()` problem. So, here it is, since it is still related to this presence.

I still don't get it what it should be a concern. This code below...

```ts
if (updateCallback.present) {
		resetData()
		updateCallback.function()
		presence.setActivity(presenceData)
	})
} else {
	...
}
function resetData() {
	currentURL = new URL(document.location.href)
	currentPath = currentURL.pathname.slice(1).split("/")
	presenceData = {
		details: <string> "Viewing an unsupported page",
		state: <string> undefined,
		largeImageKey: <string> "lg",
		startTimestamp: <number> browsingStamp,
		endTimestamp: <number> undefined
	}
}
```

...is the same as this code below,...

```ts
	currentURL = new URL(document.location.href)
	currentPath = currentURL.pathname.slice(1).split("/")
	presenceData = {
		details: <string> "Viewing an unsupported page",
		state: <string> undefined,
		largeImageKey: <string> "lg",
		startTimestamp: <number> browsingStamp,
		endTimestamp: <number> undefined
	}
	updateCallback.function()
	presence.setActivity(presenceData)
})
```

...which is also same as this code,...

```ts
	currentURL = new URL(document.location.href)
	currentPath = currentURL.pathname.slice(1).split("/")
	presenceData = {
		details: <string> "Viewing an unsupported page",
		state: <string> undefined,
		largeImageKey: <string> "lg",
		startTimestamp: <number> browsingStamp,
		endTimestamp: <number> undefined
	}

	/*

	Insert all the presence code here...

	*/
	
	presence.setActivity(presenceData)
})
```

...and this is the base which is applied by most presence in this repository, if not all.

Now, there is no difference between my base and the general base, except for one, which is when the presence is not updated or generally broken because I think that it is not a dynamic page, but it is. I have tested the presence with various pages before doing the pull requests to minimilize the said problem. One other cause of this problem is when the website is updated so it is dynamically updated, instead of being static pages. This happens on GeoGuessr, which I'll fix after few pull requests.

Also, I don't think my code is messy, but I rather think that this is an unusual implementation, so I understand how you generally dislike it.

Hopefully you understand about what I am talking about, and I apologize for this argument.

## Images

| Image | Link visited |
| ----- | ------------ |
| Discord_2020-04-14_22-23-58 | https://www.deviantart.com/yuumei/art/Phases-837756286 |


DeviantArt/presence.ts
```diff
@@ -0,0 +1,565 @@
...
+ */
+function resetData(): void {
+  currentURL = new URL(document.location.href);
+  currentPath = currentURL.pathname.slice(1).split("/");
+  presenceData = {
+    details: "Viewing an unsupported page" as string,
+    state: undefined as string,
+    largeImageKey: "lg" as string,
+    startTimestamp: browsingStamp as number,
+    endTimestamp: undefined as number
```
doomlerd, line 37: You still haven't fixed this, you know you can't just set null or undefined to a variable that's already in that state.

Hans5958: That's the reset function.

Bas950: Instead of resetting just don't send them..
`Delete presenceData.state`

Hans5958: Well, defining something as undefined is the same as not defining something.  
Still, guys, you guys keep resisting. I thought we're done about #1267. Fine.  
@doomlerd, hope you read my notes.

PR #1346, opened 2020-04-17
Update Aternos presence.

Allvaa:
- fix "Home Page" text on presence when it is on the panel page
- change "Configuring Server" to "Panel"
image


doomlerd:
Why did you remove TypeScript files?

Allvaa:
> Why did you remove TypeScript files?

Sometimes I forget to compile it, so i use js instead

doomlerd:
> Sometimes I forget to compile it, so i use js instead

Just revert to TS, we'll remind you if you forget it.

Allvaa:
ok done. could this merged right now?

PR #1350, opened 2020-04-17
RemDB - Online count & contributor.

Dekirai:
One last update for now, I hope. :)


Dekirai:
> Why did you change the author's name?

Because I want the name to be unique.
My name here is Dekirai, so I want it to be Dekirai too as Author.

Kevin could be anyone in the world, I didn't like it that way.

PR #1367, opened 2020-04-20
Rich Presence for Terminal 00

PnfrlEnm:
Sol onset.


frozaum:
WORKING
image

doomlerd:
Hello @PnfrlEnm, there has been a technical issue on our end which required some temporary changes that we will make in order for the Presence to appear in the store.

PR #1386, opened 2020-04-23
Instant Gaming support.

Dekirai:
Website: [Instant Gaming](https://www.instant-gaming.com/en/)
Preview of the presence: [Preview](https://kevin.s-ul.eu/BUYrGkQk.mp4)

There isn't really much code in it since the website works a little different because it uses different URLs for each language.
The most important stuff, see what product you are looking at, is working on every language.
When viewing a product, it will always show the platform it is for, the product's title and the product's price.

I'd take any suggestions related to the website that also could be added without any problems.
So far only user profiles and products are working.

### DIFF
--- a/Instant-Gaming/dist/metadata.json
+++ b/Instant-Gaming/dist/metadata.json
@@ -10,7 +10,7 @@
   },
   "service": "Instant-Gaming",
   "version": "1.0.0",
-  "logo": "https://i.imgur.com/ODMBHzO.png",
+  "logo": "https://i.imgur.com/ULmYGzW.png",
   "thumbnail": "https://i.imgur.com/N0SrNHb.jpg",
   "color": "#e37d00",
   "tags": [
--- a/Instant-Gaming/dist/metadata.json
+++ b/Instant-Gaming/dist/metadata.json
@@ -1,0 +1,21 @@
+{
+  "author": {
+    "name": "Dekirai",
+    "id": "601560798409654292"
+  },
+  "url": "www.instant-gaming.com",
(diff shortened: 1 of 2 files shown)


Instant-Gaming/dist/metadata.json
```diff
@@ -0,0 +1,21 @@
...
+  "author": {
+    "name": "Dekirai",
+    "id": "601560798409654292"
+  },
+  "url": "www.instant-gaming.com",
+  "description": {
+    "en": "Instant-Gaming.com - All your favourites games for Steam, Origin, Battle.net, Uplay and Indie games up to 70% off! Digital games, Instant delivery 24/7!",
+    "de": "Instant-Gaming.com - Deine Lieblingsspiele für Steam, Origin, Battle.net, Uplay und Indie Games mit bis zu 70% Rabatt! Digitaler Download, Sofortige Lieferung, 24/7!"
+  },
+  "service": "Instant Gaming",
```
Bas950, line 11: service name must be the same as folder name

Bas950: or folder name must be the same as service name

Instant-Gaming/dist/metadata.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "1.0.0",
+  "logo": "https://i.imgur.com/ODMBHzO.png",
```
frozaum, line 13: Please get a higher quality version of the logo or use [waifu2x](http://waifu2x.udp.jp/index.pt.html)

PR #1453, opened 2020-04-30
add Complex Radio

TylerCousins:


Complex Radio/presence.ts
```diff
@@ -0,0 +1,26 @@
...
+    const data = await window
+        .fetch("https://radio.complexr.pw/api/nowplaying/1")
+        .then((res) => res.json());
+    ssong = data.now_playing.song.text;
+    spresenter = data.live.is_live ? "Presenter " + data.live.streamer_name : "AutoDJ";
+    slisteners = data.listeners.unique + " Listeners";
+}
+
+setInterval(newStats, 2500);
+newStats();
```
Alanexei, line 15: You are calling an async function without await. That's the problem.

TylerCousins: Right how would i fix this. As it seems to be fine when i load the presence

Alanexei: It's part of the reason that you are getting the DeepScan issue.

TylerCousins: https://i.imgur.com/LevJzfA.png

TylerCousins: still dk how to fix that deepscan

Alanexei, line 15: This is supposed to be `await newStats()`

Alanexei: Send the screenshot directly to github by copy and pasting it into a comment.

TylerCousins: done that still issue with the deepsync

PR #1462, opened 2020-04-30
Add Edgenuity Presence

SnoweyFoxx:
image
image
image


SnoweyFoxx:
Settings:
[With Site Info]
image
[Without Site Info]
image
[With Class Info]
image
[Without Class Info]
image

PR #1467, opened 2020-05-01
🐛 DuckDuckGo - Fix Message Inconsitency

Cairo2k18:
One of them had an exclamation mark at the end and the other didn't.

(forgot to do this to DuckDuckGo and only did it to Bing #1457)


Cairo2k18:
Oops

PR #1495, opened 2020-05-04
Champion.gg Fixed issues

drull1000:
initial page
matchup
runes
statistics
Fixed some presence issues


Championgg/dist/presence.js
```diff
@@ -11,22 +11,26 @@ presence.on("UpdateData", async () => {
...
         presenceData.state = "Preparing to fight";
         presenceData.startTimestamp = time;
+        presence.setActivity(presenceData);
     }
     else if (path.startsWith("/champion")) {
         const name = path.replace("/champion/", "");
         presenceData.details = "Checking Runes";
         presenceData.state = name;
         presenceData.startTimestamp = time;
+        presence.setActivity(presenceData);
```
Bas950, line 33: I know why you needed to do these additions, (See second comment)

Championgg/dist/presence.js
```diff
@@ -11,22 +11,26 @@ presence.on("UpdateData", async () => {
...
     }
     else if (presenceData.details == null) {
```
Bas950, line 35: else if (presenceData.details == null) {

This shouldn't be else if, it should be just a standalone if

So change it to that and remove the setactivitys you added and see if it works then

drull1000: Code has been corrected, thank you for the tips :D

PR #1502, opened 2020-05-06
Change F@H url to default client

i-am-gizm0:
The F@H control opens https://client.foldingathome.org/ by default, so the presence should use that URL for ease-of-use
image


doomlerd:
Are you sure localhost/127.0.0.1 won't be used by default?

i-am-gizm0:
I haven't changed any settings on my client and that's the URL that it opens (also I don't know if it's possible to)

PR #1524, opened 2020-05-09
Structure changes

Bas950:


Bas950:
All presences are now in TypeScript
Only 1 deleted presence: Radio.net
@ACertainCoder is recoding it.

doomlerd:
Only the QA check is left.

PR #1562, opened 2020-05-19
⬆️ Updates and fix bilibili

Kotoki1337:
#1554

image
image
image
image
image

### DIFF
--- a/websites/B/bilibili/presence.ts
+++ b/websites/B/bilibili/presence.ts
@@ -1,11 +1,3 @@
-var presence = new Presence({
-    clientId: "639591760791732224"
-  }),
-  strings = presence.getStrings({
-    play: "presence.playback.playing",
-    pause: "presence.playback.paused"
-  });
-
 /**
  * Get Timestamps
  * @param {Number} videoTime Current video time seconds
@@ -15,23 +7,26 @@ function getTimestamps(
   videoTime: number,
   videoDuration: number
 ): Array<number> {
-  var startTime = Date.now();
-  var endTime = Math.floor(startTime / 1000) - videoTime + videoDuration;
+  const startTime = Date.now(),
+    endTime = Math.floor(startTime / 1000) - videoTime + videoDuration;
   return [Math.floor(startTime / 1000), endTime];
 }
 
-var browsingStamp = Math.floor(Date.now() / 1000);
-
-var user: any;
-var title: any;
-var UID: any;
-var page: any;
-var currentTime: any,
+const presence = new Presence({ clientId: "639591760791732224" }),
+  strings = presence.getStrings({
+    play: "presence.playback.playing",
+    pause: "presence.playback.paused"
+  });
+const browsingStamp = Math.floor(Date.now() / 1000);
+let user: any, title: any, inread_title: any, UID: any, page: any;
+let video: any, videoDuration: any, videoCurrentTime: any, videoPaused: any;
+const huodong_title = document.querySelector("#viewbox_report > h1 > a"),
+  hudong_title = document.querySelector("#viewbox_report > h1 > span.activity");
+let currentTime: any,
   duration: any,
   paused: any,
   playback: any,
   iFramePaused: any;
-
 presence.on("iFrameData", (data) => {
   playback = data.iframe_video.duration !== null ? true : false;
 
@@ -47,6 +42,7 @@ presence.on("UpdateData", async () => {
     largeImageKey: "bb"
   };
 
+  // Maincode
   if (document.location.hostname == "www.bilibili.com") {
     if (document.location.pathname == "/") {
       presenceData.startTimestamp = browsingStamp;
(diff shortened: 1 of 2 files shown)


websites/B/bilibili/presence.ts
```diff
@@ -103,37 +103,50 @@ presence.on("UpdateData", async () => {
...
+      }
       presenceData.details = title.innerText;
-      presenceData.state = "By user: " + user.innerText;
+      presenceData.state = user.innerText;
+
       //播放历史
     } else if (document.location.pathname == "/account/history") {
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Viewing history";
+      presenceData.details = "Their history";
```
Bas950, line 130: `Viewing their history` might be better
Viewing history or Their history is both not really good.

websites/B/bilibili/presence.ts
```diff
@@ -720,13 +286,12 @@ presence.on("UpdateData", async () => {
         "#app > div.page-container.p-rel > div.top-header.p-rel > div.tag-title-content.fs-28.ls-0 > div.tag-title.d-i-block"
       );
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Browsing dynamic";
+      presenceData.details = "Dynamic";
       presenceData.state = "Tag: " + title.innerText;
     } else {
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Browsing for dynamic";
+      presenceData.details = "Their dynamic";
```
Bas950, line 293: Their dynamic doesn't sound correct.

websites/B/bilibili/presence.ts
```diff
@@ -219,475 +232,28 @@ presence.on("UpdateData", async () => {
...
-      presenceData.details = "Viewing TV series";
-      presenceData.state = "└ China domestic";
-    } else if (document.location.pathname.includes("/v/tv/overseas/")) {
-      presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Viewing TV series";
-      presenceData.state = "└ China overseas";
-    } else if (document.location.pathname.includes("/tv/index/")) {
-      presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Viewing TV series";
-      presenceData.state = "└ index";
```
Bas950, line 690: Why was this all removed?

websites/B/bilibili/presence.ts
```diff
@@ -764,27 +329,24 @@ presence.on("UpdateData", async () => {
...
+      presenceData.state = user.innerText;
     } else {
       presenceData.startTimestamp = browsingStamp;
       presenceData.details = "Browsing Someone's space";
     }
   } else if (document.location.hostname == "account.bilibili.com") {
     if (document.location.pathname.includes("/")) {
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Viewing account setting";
+      presenceData.details = "Their account settings";
```
Bas950, line 343: `Viewing their settings` might be better

websites/B/bilibili/presence.ts
```diff
@@ -764,27 +329,24 @@ presence.on("UpdateData", async () => {
...
     }
     //创作中心
   } else if (document.location.hostname == "member.bilibili.com") {
     if (document.location.pathname.includes("/")) {
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Viewing 创作中心";
+      presenceData.details = "Their dashboard";
```
Bas950, line 349: `Viewing their dashboard`

websites/B/bilibili/presence.ts
```diff
@@ -843,60 +412,61 @@ presence.on("UpdateData", async () => {
...
       );
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Browsing for live category";
-      presenceData.state = "└ " + title.innerText;
+      presenceData.details = "Viewing live channel";
+      presenceData.state = title.innerText;
     } else {
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Browsing for live channels";
+      presenceData.details = "Viewing live channel";
```
Bas950, line 429: channels* you forgot an s at the end

websites/B/bilibili/presence.ts
```diff
@@ -843,60 +412,61 @@ presence.on("UpdateData", async () => {
...
         " - 搜索结果 - 哔哩哔哩弹幕视频网 - ( ゜- ゜)つロ  乾杯~  - bilibili",
         ""
       );
+      presenceData.smallImageKey = "search";
     }
   } else if (document.location.hostname == "message.bilibili.com") {
     if (document.URL.includes("/#/reply")) {
       presenceData.startTimestamp = browsingStamp;
-      presenceData.details = "Viewing message";
+      presenceData.details = "Their message";
```
Bas950, line 445: their message sounds kinda strange aswell

websites/B/bilibili/presence.ts
Kotoki1337: > `Viewing their history` might be better
> Viewing history or Their history is both not really good.

i saw that youtube uses "their history" so i copied it

websites/B/bilibili/presence.ts
Kotoki1337: > Why was this all removed?

these things are very deep, and no one will go into these places. i am thinking of other more concise ways to implement them. some places that no one visits and take up too much code block. i think it is a bit too messy.

Bas950: If it works then there is no reason to remove it, even if close to noone goes there.

Kotoki1337: > If it works then there is no reason to remove it, even if close to noone goes there.

i implemented it in a more concise way

PR #1565, opened 2020-05-19
Optimized SoundCloud and Rewrote Mixer

Alanexei:
Did not include screenshots for SoundCloud as it did not affect anything visually. Only modified the queries and other things.

**Mixer**:
image
image


Alanexei:
image
image

PR #1566, opened 2020-05-19
New Presence Methods

Alanexei:
Added new presence methods.

**Do not merge until extension v2.1.1 has been released to the public.**


LewisTehMinerz:
Types are wrong, it needs to be `iFrameRegExp` rather than `iframeRegExp`. `iframe`, however, is correct.

Alanexei:
This got all messed up. Remaking this.

PR #1567, opened 2020-05-19
Changes²

doomlerd:


doomlerd:
@PreMiD/presence-verifiers please make sure to review the last commit carefully with attention to any change that is not for a type.

PR #1580, opened 2020-05-21
add schema validation system

LewisTehMinerz:


doomlerd:
I don't think that's a very useful way, we can't force devs to use a schema and we can't rely on them doing so either.

LewisTehMinerz:
@doomlerd I already anticipated that, that's why `$schema` being missing throws a warning rather than a full error.

doomlerd:
Then it's of no use.

LewisTehMinerz:
Just FYI, `$schema` provides autocomplete:

and required properties:


We don't need to force the devs to use it, but people may opt in to using it. Having this validation system on the CI means that, even if they opt out, it'll still validate their metadata.

doomlerd:
Idk seems forced to me, and even conflicts with the "incoming" changes I was planning. I guess it was clear when I said that was my task but you decided to still ignore that anyways, let's hope it works.

PR #1616, opened 2020-05-25
Remove Schema from Metadata

Alanexei:


doomlerd:
would be better if you use "delete", and did you make sure it works? xD

Alanexei:
If I use delete it would probably throw an error if they key doesn't exit. It's only one line though so it should work.

PR #1625, opened 2020-05-26
Updated presence.ts query selector of video

kamui-7:
The class name of the video element has changed to `videoo1_html_api`


websites/0-9/4anime/presence.ts
```diff
@@ -49,7 +49,7 @@ presence.on("UpdateData", async () => {
   }
 
   var video: HTMLVideoElement =
```
Gurrman375, line 51: Change to `const video: HTMLVideoElement =`

doomlerd:
already fixed #1648

PR #1653, opened 2020-05-30
SnowBoltz Bugs Fixed

Yusaki-Akito:
[-] Login state
[+] Home page path fixed
[+] Fixed weird bugs that not work for some people
[+] Add something on website change a litle bit
image


Yusaki-Akito:
Unknow error lol

PR #1655, opened 2020-05-30
Update presence.ts

Limerio:
Add API modules exist in nodejs


frozaum:
Going to close this in a day.

PR #1778, opened 2020-06-20
New Presence (SuperAnimes)

Nanashinn:
image
image
image


websites/S/Super Animes/dist/metadata.json
```diff
@@ -0,0 +1,20 @@
+{
+  "author": {
+    "name": "Nanashin",
+    "id": "330030557457022986"
+  },
+  "service": "Super Animes",
+  "description": {
+    "en": "",
```
eggsywashere, line 8: Don't leave the descriptions empty....

websites/S/Super Animes/iframe.ts
```diff
@@ -0,0 +1,81 @@
...
+            iframe.send ({
+                currentTime: video.currentTime,
+                duration: video.duration,
+                played,
+                paused: video.paused
+            });
+        }
+    }
+
+    if (document.querySelector("#video") != null) {
```
eggsywashere, line 30: You don't have to do it like this all the time. `querySelector` won't throw an error if it couldn't find the element. It'd return `null` so you can use the or operator (`||`) without any problem. This will save you from the unnecessary code and will beautify your code. Example:

```ts
const video: HTMLVideoElement = document.querySelector("#play0") || document.querySelector("#video")

if (video && !isNaN(video.currentTime) iframe.send({});
```

websites/S/Super Animes/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "../tsconfig.json",
+    "compilerOptions": {
+      "outDir": "dist"
+    }
+  }
```
Bas950, line 6: https://docs.premid.app/en/dev/presence#filling-in-the-tsconfigjson-file

websites/S/Super Animes/iframe.ts
doomlerd: or just 
`const video: HTMLVideoElement = document.querySelector("video")`

nenilves:
mano, te amo <3 
tava doido por essa presence novamente

PR #1785, opened 2020-06-22
Some fix (Super Animes)

Nanashinn:


Nanashinn:
> show proof of it working

image
image

PR #1804, opened 2020-06-28
WhatsApp Fix

GregTCLTK:
Updated the names of some div classes so the presence can work again


GregTCLTK:
Before:
image

After:
image

PR #1828, opened 2020-07-05
Add Formuliser presence

Oj18:
Adds presence for https://formuliser.oojmed.com

### DIFF
--- a/websites/F/Formuliser/dist/metadata.json
+++ b/websites/F/Formuliser/dist/metadata.json
@@ -11,7 +11,7 @@
   "service": "Formuliser",
   "version": "1.0.0",
   "logo": "https://i.imgur.com/rtkOdlB.png",
-  "thumbnail": "http://i.imgur.com/nEJz6j6.png",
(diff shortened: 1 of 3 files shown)


Oj18:
Here it is - made another commit which fixed a bug and should of also fixed DeepScan:
image

websites/F/Formuliser/dist/metadata.json
```diff
@@ -0,0 +1,20 @@
...
+    "id": "506482395269169153"
+  },
+  "url": "formuliser.oojmed.com",
+  "description": {
+    "en": "Get information on chemical formulae."
+  },
+  "service": "Formuliser",
+  "version": "1.0.0",
+  "logo": "https://i.imgur.com/rtkOdlB.png",
+  "thumbnail": "http://i.imgur.com/nEJz6j6.png",
```
Alanexei, line 14: This needs to be HTTPS.

Oj18: Weird it isn't, fixing it in new commit now.

PR #1856, opened 2020-07-12
Updated ChickenFM Presence

TheChicken14:
Hi, I updated the ChickenFM Presence for my new website


TheChicken14:
Here's proof that it does work:
image

PR #1886, opened 2020-07-21
Update Twitch and Rewrite SoundCloud and SoloLearn

Alanexei:
Twitch now has a new setting to turn off timestamps, SoundCloud and SoloLearn received a new look.

<details>
  <summary>Twitch</summary>
  
 ## **Settings**
image

## **Preview**
image

</details>

<details>
<summary>SoundCloud</summary>

## **Settings**
image

## **Website**
image

## **Preview**
image

</details>

<details>
<summary>SoloLearn</summary>

## **Settings**
image

## **Website**
image

## **Preview**
image

</details>


websites/T/Twitch/presence.ts
```diff
@@ -297,6 +292,10 @@ presence.on("UpdateData", async () => {
       data.smallImageKey = "reading";
       data.smallImageText = (await strings).browse;
     }
+    if (!showTimestamps) {
+      data.startTimestamp = undefined;
+      data.endTimestamp = undefined;
+    }
```
Bas950, line 298: `delete data.startTimestamp` instead of setting it to undefined.

websites/S/SoundCloud/presence.ts
```diff
@@ -1,84 +1,278 @@
...
+      data.smallImageText = (await strings).search;
+    }
+    if (data.details.match("(Uploading)")) {
+      data.smallImageKey = "uploading";
+      data.smallImageText = "Uploading..."; // no string available
+    }
+    if (!showTimestamps || (!playing && !showBrowsing)) {
+      data.startTimestamp = undefined;
+      data.endTimestamp = undefined;
+    }
```
Bas950, line 271: `delete data.startTimestamp` instead of setting it to undefined.

websites/S/SoloLearn/presence.ts
```diff
@@ -1,174 +1,164 @@
...
-
-  if (path.match("/Blog")) {
-    data.details = "Viewing Blog";
-
-    const name = getElement(".articleTitle");
-    if (name) {
-      data.state = name;
+    if (!showTimestamps) {
+      data.startTimestamp = undefined;
+      data.endTimestamp = undefined;
```
Bas950, line 160: `delete data.startTimestamp` instead of setting it to undefined.

Alanexei: Deleting is not efficient because it changes the structure of an object, while setting it to undefined releases the content but doesn't force the object to modify its structure.

Bas950: IMO you shouldnt send anything that doesnt need sending.

doomlerd: @Alanexei Actually, when you set it to undefined it is never deleted and it's not the same as if the value was never set. Sure it does modify the structure and that's the point, because the property is better off destroyed than just modified.
While it may seem illogical, it is mostly Discord to be blamed since I guess they check for the type in some different way that still keeps the value defined even when you set it to `undefined`, same applies to `null` so delete is the best choice, we could find a workaround on the extension level (as usual) of course but for now that's how it is.

doomlerd: Also to back what Bas said. Since it holds value when just set as undefined, you're still sending something which slightly affects performance.

Alanexei: My point is to not delete the value, it is to reset the value. It doesn't really matter either way, because the presence behaves as it should. But I will change it to not cause any more fuss. If you are enforcing the `delete` keyword, you should have asked me to put it in the guidelines.

doomlerd: We've discussed it before.

PR #1896, opened 2020-07-23
Shazam Presence

Alanexei:
Shazam presence that fully supports the website.

image
image


websites/S/Shazam/presence.ts
```diff
@@ -0,0 +1,128 @@
...
+};
+
+let elapsed = Math.floor(Date.now() / 1000),
+  prevUrl = document.location.href;
+
+const statics = {
+  "/": {
+    details: "Browsing...",
+    state: "Home"
+  },
```
doomlerd, line 30: good one

PR #1897, opened 2020-07-23
Code maintenance

doomlerd:
- Changes ESLint config and its override to match the current ECMAScript target of our TypeScript config;
- Disallows empty lines at the start/end of files and sets the maximum of empty lines between functions to 1 (self explanatory);
- "Allows multiple variable declarations per scope but requires consecutive variable declarations to be combined into a single declaration" (not much of a performance difference, reduces file sizes);
- Forces linebreak style to LF ([why](https://jessitron.com/2019/11/11/line-endings-in-git));
- `noInlineConfig` - disables inline configuration of ESLint for security purposes;
- `reportUnusedDisableDirectives` - self explanatory, related to the previous rule;
- Overrides the rules for non-Presence scripts (e.g. the updater needs a node environment compared to Presences which need a browser environment) and allows non-null-assertion (for the TypeScript compiler)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,7 @@
   "root": true,
   "env": {
     "browser": true,
-    "es6": true
+    "es2020": true
   },
   "extends": [
     "eslint:recommended",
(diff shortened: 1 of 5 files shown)


doomlerd:
wut

PR #1960, opened 2020-08-12
add default setting

paz:


paz:
here

doomlerd:
Update your fork, it's 50 commits behind
and fix the issues thrown by DeepScan afterwards

PR #1965, opened 2020-08-14
Edited DualFM - Removed Elapse Toggle

jackkelly-bayliss:
I have removed the ability to toggle the 'elapse' status due to issues with first installation of the extension. Thank you.


websites/D/DualFM/presence.ts
```diff
@@ -31,22 +31,16 @@ presence.on("UpdateData", async () => {
     largeImageKey: "logo",
     smallImageKey: "dualfm-play-v2"
   };
-
-  const toggleelaspe = await presence.getSetting("toggleelapse");
   const changedetails = await presence.getSetting("changedetails");
   const changestate = await presence.getSetting("changestate");
   const changesmalltext = await presence.getSetting("changesmalltext");
```
Bas950, line 36: instead of removing it, give it a default false:
const toggleelaspe = await presence.getSetting("toggleelapse") || true;

Bas950: do that for all of them btw

jackkelly-bayliss: I didn't even think about adding that, you're a genius haha. Thank you :)

websites/D/DualFM/dist/metadata.json
```diff
@@ -23,12 +23,6 @@
   ],
   "category": "music",
   "settings": [
-    {
-      "id": "toggleelapse",
-      "title": "Toggle Elapse Status",
-      "icon": "fas fa-history",
-      "value": true
-    },
     {
```
Bas950, line 26: undo

websites/D/DualFM/presence.ts
```diff
@@ -32,7 +32,7 @@ presence.on("UpdateData", async () => {
     smallImageKey: "dualfm-play-v2"
   };
 
-  const toggleelaspe = await presence.getSetting("toggleelapse");
+  const toggleelaspe = await presence.getSetting("toggleelapse") || true;
   const changedetails = await presence.getSetting("changedetails");
   const changestate = await presence.getSetting("changestate");
   const changesmalltext = await presence.getSetting("changesmalltext");
```
Bas950, line 38: give default value

jackkelly-bayliss: They already have a default value below, or is that irrelevant and I still need to define it in the variable itself?

Bas950: They will give the same error probs as what the first one did, so just do it to be sure.
Btw to fix the error you just have to change a setting (turn off and back on)

Bas950: update your fork too

jackkelly-bayliss: Updated the fork and that's what we found, we found that toggling the 'Toggle Elapse' setting seemed to fix it but it's a little inconvenient for some users haha. I've just added the some default values to the variables, hopefully all will be okay. Thank you for your assistance Bas and feel free to let me know if there's anything else that needs changing.

doomlerd:
5 more issues to fix.

websites/D/DualFM/dist/metadata.json
```diff
@@ -3,12 +3,12 @@
     "name": "Jack Hansen",
     "id": "140590672657776641"
   },
-  "service": "DualFM",
+  "service": "Dual",
```
Bas950, line 6: service name and folder name must be the same. Note: if you change the service name rn everyone would need to re-add the presence

Bas950 (changes requested):

websites/D/DualFM/dist/metadata.json
jackkelly-bayliss: Thank you for the notice, sadly - I think to prevent confusion, we need to rename the service name to match all our branding otherwise we know people will somehow get confused haha. Thank you Bas.

Bas950: Aight then you also have to change the folder name to the new one

Bas950 (approved):

websites/D/Dual/presence.ts
```diff
@@ -0,0 +1,88 @@
...
+	lastTimeStart = Math.floor(Date.now() / 1000);
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "logo",
+    smallImageKey: "dualfm-play-v2"
+  },
+  changedetails = await presence.getSetting("changedetails"),
+  changestate = await presence.getSetting("changestate"),
+  changesmalltext = await presence.getSetting("changesmalltext");
```
Bas950, line 36: the `|| true` 's are gone?

Bas950 (changes requested):

DamonOnYT (approved):
i approve good work jack now please respond to my dms

doomlerd:
Closing in one day.

websites/D/Dual/presence.ts
jackkelly-bayliss: I don’t personally think adding the fallback values for those values are necessary as we’ve not received any issues with those setting values. The only issues we’ve received are regarding the ‘Elapse’ setting.

Bas950 (approved):

PR #2012, opened 2020-08-24
Remove KissAnime presence

Hans5958:
## Preword

I heard KissAnime and KissManga is gone forever, and I saw that discontinued services like Mixer got their presence removed here. So, why don't we also remove KissAnime?

## Notes

- This pull request removes KissAnime from the store.


Isa27-0:
Yep, Not only this, but also the Club Penguim Rewritten. Probably shut down the servers because Disney put copyright now (It's a long story) 
I also wanted to create an Issue just to delete it, but as you already mentioned something similar, I ended up mentioning here

Hans5958:
> 
> 
> Yep, Not only this, but also the Club Penguim Rewritten. Probably shut down the servers because Disney put copyright now (It's a long story)
> I also wanted to create an Issue just to delete it, but as you already mentioned something similar, I ended up mentioning here

You sure about CPRewritten? I remember that it ever got shut down, but then it got back, until this day. Maybe you confuse this with other Club Penguin Private Servers. For history purposes, the currently ongoing event is "[Water Party](https://twitter.com/CPRewritten/status/1296511796426223622?s=20)," which ends in 27 August 2020.

Isa27-0:
Oh sorry. I just thought they would disable it, since Copyright is complicated for some

Bas950 (approved):

fr0r:
Hello everyone, I would approve this pull request if it wasn't the fact that one of the URLs is still around, even though it's not operated by Kiss themselves and just someone who probably wants to profit with the name (or probably scam people), we had a debate whether if we're going to let the presence up or not and the decision was to stay the presence on the store (which I'm still against it for obvious reasons)

I will bring this PR to our Presence Verifiers and see what they're going to say.

fr0r:
After proper discussion, we decided to approve the pull request and we will (soon) discuss about guidelines changes if something like that happens again.

Hans5958:
> 
> 
> Hello everyone, I would approve this pull request if it wasn't the fact that one of the URLs is still around, even though it's not operated by Kiss themselves and just someone who probably wants to profit with the name (or probably scam people), we had a debate whether if we're going to let the presence up or not and the decision was to stay the presence on the store (which I'm still against it for obvious reasons)
> 
> I will bring this PR to our Presence Verifiers and see what they're going to say.

Most of the time the domain is already parked, then parking domain "companies" show stuff that is not related with the original content, so that's why you often see that awkward "Privacy Policy" or "Copyright" texts, or just spam.

Hans5958:
@fr0r

PR #2025, opened 2020-08-26
Fix Crowdin Presence

TheDropX:
image


Bas950 (changes requested):
Fix deepscan errors

TheDropX:
my deepscan extension stopped working for some reason (in case you wonder why i commited so many times 🤣)

Bas950 (approved):

PR #2086, opened 2020-09-03
Remove Google Classroom presence

Allvaa:
i don't want to maintain it again and might have a lot of bug in the future if i don't maintain it anymore. so please remove the presence


kawakami:
If it is working, it can stay up. If it would need update someone probably will do it.

Allvaa:
i just want the presence to be removed. can i?

kawakami:
It shouldnt be deleted because site is working presence too. You dont need to be dev of it. If it will break or something. Someone will change you to his id

Allvaa:
ok then

PR #2095, opened 2020-09-03
Fix Aternos presence

Allvaa:
fix that the presence text is "Home Page" even when the user opens the panel


Bas950 (changes requested):
upload proof of it working

Allvaa:
image
image
image

doomlerd (approved):

Bas950 (approved):

PR #2177, opened 2020-09-20
Add Fresh Radio Presence

zoeyhockey:


zoeyhockey:
- I've done plenty of test on the presence it works like a beauty! 
- The fetch functions are at an interval of 5 seconds (there getting called every 5 seconds)

Bas950:
Website isn't old enough, it needs to be 2 months old.

PR #2209, opened 2020-09-24
Update presence.ts

MicroBotCodes:
https://user-images.githubusercontent.com/48334228/92813077-0b717500-f3b0-11ea-914d-20aecc5cf4d5.png
We got our own azuracast for beastfm so it wont load if I wont change it.


Bas950 (changes requested):
bump version

PR #2218, opened 2020-09-25
SySAlfie.xyz Request

alfiefroggoman:
**Metadata.json:**

{
    "$schema": "https://schemas.premid.app/metadata/1.0",
    "author": {
      "name": "Alfie",
      "id": "749558323795329044"
    },
    "url": "sysalfie.xyz",
    "description": {
      "english": "SysAlfie is a webpage about MrAlfieGamer & his discord bots."
    },
    "service": "SYS Alfie",
    "version": "1.0",
    "logo": "https://imgur.com/e4TH8pe",
    "thumbnail": "https://imgur.com/e4TH8pe",
    "color": "#FF0000",
    "tags": [
      "Developer",
      "Website"
    ],
    "category": "other"
  }

**Tsconfig.json**:

{
    "extends": "../../../tsconfig.json",
    "compilerOptions": {
      "outDir": "./dist/"
    }
  }

**Presence.ts**: 

var presence = new Presence({
    clientId: "759052972927483975" //The client ID of the Application created at https://discordapp.com/developers/applications
}),

    play: "presence.playback.playing",
    pause: "presence.playback.paused"
    //You can use this to get translated strings
});

/*

function myOutsideHeavyLiftingFunction(){
    //Grab and process all your data here

    // element grabs //
    // api calls //
    // variable sets //
}

setInterval(myOutsideHeavyLiftingFunction, 10000);
//Run the function separate from the UpdateData event every 10 seconds to get and set the variables which UpdateData picks up

*/

    /*UpdateData is always firing, and therefore should be used as your refresh cycle, or `tick`. This is called several times a second where possible.

    It is recommended to set up another function outside of this event function which will change variable values and do the heavy lifting if you call data from an API.*/

    var presenceData: presenceData = {
        largeImageKey: "key", /*The key (file name) of the Large Image on the presence. These are uploaded and named in the Rich Presence section of your application, called Art Assets*/
        smallImageKey: "key", /*The key (file name) of the Large Image on the presence. These are uploaded and named in the Rich Presence section of your application, called Art Assets*/
        smallImageText: "Some hover text", //The text which is displayed when hovering over the small image
        details: "Browsing Page Name", //The upper section of the presence text
        state: "Reading section A", //The lower section of the presence text
        startTimestamp: 1577232000, //The unix epoch timestamp for when to start counting from
        endTimestamp: 1577151472000 //If you want to show Time Left instead of Elapsed, this is the unix epoch timestamp at which the timer ends
    }; /*Optionally you can set a largeImageKey here and change the rest as variable subproperties, for example presenceSata.type = "blahblah"; type examples: details, state, etc.*/

    if (presenceData.details == null) {
        //This will fire if you do not set presence details
        presence.setTrayTitle(); //Clears the tray title for mac users
        presence.setActivity(); /*Update the presence with no data, therefore clearing it and making the large image the Discord Application icon, and the text the Discord Application name*/
    } else {
        //This will fire if you set presence details
        presence.setActivity(presenceData); //Update the presence with all the values from the presenceData object
    }
});

### DIFF
--- a/websites/S/SoundCloud/dist/metadata.json
+++ b/websites/S/SoundCloud/dist/metadata.json
@@ -15,7 +15,7 @@
     "zh_TW": "SoundCloud是一個線上音樂分享平臺，它允許人們合作、交流和分享原創音樂錄音。"
   },
   "url": "soundcloud.com",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "logo": "https://i.imgur.com/KgOkfIz.png",
   "thumbnail": "https://i.imgur.com/eZyrvbS.jpg",
   "color": "#FF7E30",
--- a/websites/S/SoundCloud/presence.ts
+++ b/websites/S/SoundCloud/presence.ts
@@ -152,8 +152,8 @@ presence.on("UpdateData", async () => {
     elapsed = Math.floor(Date.now() / 1000);
   }
 
-  const playButton = document.querySelector(".playControls__play.playing");
-  const playing = playButton ? true : false;
+  const playButton = document.querySelector(".playControls__play.playing"),
+        playing = playButton ? true : false;
 
   if ((playing || (!playing && !showBrowsing)) && showSong) { // deepscan-disable-line
     data.details = getElement(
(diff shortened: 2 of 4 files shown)


Bas950:
ye... no... to many issues with this PR

Issue #1734, opened 2020-06-11
Padlet | padlet.com


geenva:
**Prerequisites and essential questions** 
- [X] Is it a popular site?
- [X] Is the website older than 2 months? 
- [ ] Is the site locked to a specific country/region?
- [ ] Is the site a paid service? (e.g. Netflix, Hulu)
- [ ] Does the website feature NSFW content? (e.g. porn, etc...)
- [ ] Are you a donator/patron?
- [x] Do you acknowledge that coding presences is completely voluntary and may take time for your service to be added regardless of priority?

**What's your Discord username?** geneva#0058 

**What's the name of the service?** Padlet 

**What should the Presence display?** Padlet, Padlet name and time used to view Padlet 

**If possible, please provide a logo for the service (512x512 minimum)**  
![image](https://user-images.githubusercontent.com/34730060/84370086-58577b00-ac0a-11ea-9c31-53ce0064dfb7.png)

Mirazex:
I will do it

flarely:
Thank you!

PR #2257, opened 2020-09-30
New Presence: Padlet

Mirazex:
added new presence (simple version) for padlet.com
Issue: #1734

Proof:
2wlLMyvr1N

### DIFF
--- a/websites/P/Padlet/presence.ts
+++ b/websites/P/Padlet/presence.ts
@@ -1,4 +1,4 @@
-var presence = new Presence({
+const presence = new Presence({
   clientId: "760981805695762441"
 });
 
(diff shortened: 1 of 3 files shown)


websites/P/Padlet/presence.ts
```diff
@@ -0,0 +1,44 @@
...
+  isPadlet = !!PadletElement;
+
+  if (isPadlet) {
+    data.details = PadletTitle;
+    data.state = 'Working with Padlet...';
+    data.startTimestamp = getTimeStamp();
+  } else {
+    data.details = 'Idle';
+    data.state = 'Idle';
+  }
```
Bas950, line 41: 2 statuses aint enough

Bas950 (changes requested):

websites/P/Padlet/presence.ts
Mirazex: ok, I don't know what else to add there yet)
think of something later

Bas950:
Has been 1 week since last change, if you are going to fix it create a new PR...

PR #2270, opened 2020-10-03
Updated metadata.json of Amazon Music

STK101:
Added url to the Indian servers of Amazon music.


Bas950:
It should already work due to the regex.

PR #2312, opened 2020-10-12
Fix Hulu video player element

kagbogre:
Fixes Hulu’s presence. The video element it’s currently looking for is no longer correct.

Before fix:
image

After fix:
image


Bas950 (changes requested):
bump version

Bas950 (changes requested):
it isn't that big of a change to be able to add yourself as contributor of the presence.

Bas950 (approved):

Timeraa (approved):

PR #2361, opened 2020-10-20
Add BigBlueButton presence v1.0.0.

kuba2k2:
Some examples of the presence working:
screenshot

### DIFF
--- a/websites/B/BigBlueButton/dist/metadata.json
+++ b/websites/B/BigBlueButton/dist/metadata.json
@@ -1,0 +1,19 @@
+{
+    "$schema": "https://schemas.premid.app/metadata/1.0",
+    "author": {
+        "name": "Oranż Metylowy",
+        "id": "359982590972788737"
+    },
+    "url": "demo.bigbluebutton.org",
+    "regExp": "\\/b\\/[\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client\\/join\\?",
+    "description": {
+        "en": "BigBlueButton online learning platform"
+    },
+    "service": "BigBlueButton",
+    "version": "1.0.0",
+    "logo": "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/bbb_logo.png",
+    "thumbnail": "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png",
+    "color": "#0F70D7",
+    "tags": ["bbb", "education", "learning", "voicecall"],
+    "category": "socials"
+}
--- a/websites/B/BigBlueButton/dist/metadata.json
+++ b/websites/B/BigBlueButton/dist/metadata.json
@@ -4,16 +4,32 @@
         "name": "Oranż Metylowy",
         "id": "359982590972788737"
     },
-    "url": "demo.bigbluebutton.org",
-    "regExp": "\\/b\\/[\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client\\/join\\?",
+    "url": [
+        "demo.bigbluebutton.org",
+        "demo2.bigbluebutton.org"
+    ],
+    "regExp": "\\/(b|gl)\\/[\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client\\/join\\?|client\\/guest-wait.html",
     "description": {
         "en": "BigBlueButton online learning platform"
     },
     "service": "BigBlueButton",
     "version": "1.0.0",
-    "logo": "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/bbb_logo.png",
-    "thumbnail": "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png",
+    "logo": "https://i.imgur.com/qjmxCvj.png",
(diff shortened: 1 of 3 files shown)


Bas950:
Marked as ready for review since we don't allow drafts.

websites/B/BigBlueButton/dist/metadata.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "url": "demo.bigbluebutton.org",
+    "regExp": "\\/b\\/[\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client\\/join\\?",
+    "description": {
+        "en": "BigBlueButton online learning platform"
+    },
+    "service": "BigBlueButton",
+    "version": "1.0.0",
+    "logo": "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/bbb_logo.png",
+    "thumbnail": "https://raw.githubusercontent.com/bigbluebutton/greenlight/master/app/assets/images/logo_with_text.png",
```
Bas950, line 15: upload to imgur

websites/B/BigBlueButton/dist/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
+    "$schema": "https://schemas.premid.app/metadata/1.0",
+    "author": {
+        "name": "Oranż Metylowy",
+        "id": "359982590972788737"
+    },
+    "url": "demo.bigbluebutton.org",
+    "regExp": "\\/b\\/[\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client\\/join\\?",
+    "description": {
```
Bas950, line 9: for what exactly is this regex?

websites/B/BigBlueButton/presence.ts
```diff
@@ -0,0 +1,106 @@
...
+    strings = presence.getStrings({
+        play: "presence.playback.playing",
+        pause: "presence.playback.paused"
+        //You can use this to get translated strings
+    });
+
+var roomName: string;
+var inCall: boolean;
+var userCount: number;
+var userState: string;
```
Bas950, line 14: no var use const/let
and then merge them too

Bas950 (changes requested):
Remove comments and upload proof of it working

websites/B/BigBlueButton/dist/metadata.json
kuba2k2: This is for checking the meeting URL, as BBB is mostly self-hosted so a universal domain URL is not really possible.

Bas950: Could you provide an example url for me.

kuba2k2: https://bbb.school.city.com/b/xyz-97h-jki
or after joining the session
https://bbb.school.city.com/html5client/join?sessionToken=1234567890abcdef
Will update the regexp to also match the guest lobby URLs in the next commit.

Bas950: `[/]b[/][\w]{3}-[\d\w]{3}-[\d\w]{3}|html5client[/]join[?]`
image
Make sure you use the JS version cause some stuff wasn't working.

kuba2k2: obraz
Sure, the regex works fine on both PCRE and JS implementations. The PreMiD extension also handles it correctly (thus showing correct presence, as in my first screenshot).

Bas950: We would prefer it if you use the JS flavor since its a JS script.
And instead of escaping with backslash put [] around it.

ririxichan (dismissed):

Bas950 (changes requested):
https://github.com/PreMiD/Presences/pull/2361#discussion_r509236658

websites/B/BigBlueButton/dist/metadata.json
kuba2k2: Okay, done.

websites/B/BigBlueButton/dist/metadata.json
```diff
@@ -8,7 +8,7 @@
         "demo.bigbluebutton.org",
         "demo2.bigbluebutton.org"
     ],
-    "regExp": "\\/(b|gl)\\/[\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client\\/join\\?|client\\/guest-wait.html",
+    "regExp": "[/](b|gl)[/][\\w]{3}-[\\d\\w]{3}-[\\d\\w]{3}|html5client[/]join[?]|client[/]guest-wait[.]html",
```
Bas950: I am pretty sure that in the js varient \\d is just \d same with \\w -> \w.

kuba2k2: The backslashes in \d and \w have to be escaped to be inside a JSON file. \d is not an escaped symbol for JSON, but for regex, where it already has only one backslash. Not sure if I understand  what you mean correctly.

Bas950: ah aight, did you test it if it still works then?

kuba2k2: Yes, my last commit changing backslashes to [] still works.

Bas950 (approved):

ririxichan (approved):

PR #2384, opened 2020-10-25
Fixed "Unknown page" for Games and Catalog

GRIPPERTV:
I saw that when going to "/games/" or "/catalog/", the path was different (on my Chrome). It was "/games#/" and "/catalog?Category=", and in the code checks if includes "/games/" and "/catalog/", so I just removed the "/" from the end (and worked perflecty).

The modifications are simple to resolve, but I am uncomfortable having to go to "/games" and the presence results is "Unknown page".

### DIFF
--- a/websites/R/Roblox/dist/metadata.json
+++ b/websites/R/Roblox/dist/metadata.json
@@ -14,7 +14,7 @@
     "www.roblox.com",
     "web.roblox.com"
   ],
-  "version": "2.3.5",
+  "version": "2.3.6",
   "logo": "https://i.imgur.com/O8T2o2j.png",
   "thumbnail": "https://i.imgur.com/pufi4q1.png",
   "color": "#1b1d1e",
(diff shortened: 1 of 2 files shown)


ririxichan (changes requested):
bump version also give proof of working

GRIPPERTV:
> bump version also give proof of working

image
As you can see, the first statement checks if the path includes `/games/`, and the result is false
In the second statement checks if the path includes `/games` (without the `/` at the end), and the result is true
Well, I think you can already consider that it works

After some redirect tests, the `#` is always included in the path, even if redirecting to `www.roblox.com/games`
image

ririxichan:
I meant proof by showing on presence status but ok

GRIPPERTV:
> I meant proof by showing on presence status but ok still bump version

Oh, 2.3.6

Bas950 (changes requested):
upload proof of the presence working

ririxichan:
No proofs from pull request creator. Closing.

PR #2399, opened 2020-10-26
Updated Starz Presence

kagbogre:
Modified the `title !== null` check to `if(title)` to cover all falsy values (including "").

Had problems while developing the Philo presence because the title was "", so I'm changing it here as well, just in case.


Bas950 (changes requested):
Upload proof of it working

kagbogre:
> Upload proof of it working

image
image

Bas950 (approved):

ririxichan (approved):

PR #2406, opened 2020-10-28
Updated Philo Presence

kagbogre:
Always show episode info if available (even if live)

Before:
image

Now: 
image


kagbogre:
Added channel name to presence.


Example with "on AMC":
image

kagbogre:
Updated movie string to "Watching" so it looks nicer with the new channel data

image

instead of "Movie on AMC"

kagbogre:
image

Proof that it still works after last change. That’s all I’ll be including in this PR.

ririxichan (approved):

Bas950 (approved):

PR #2410, opened 2020-10-31
Add russian lang to juniperbot presence

MrLivixx:


Bas950:
We are releasing presence languages soon, so wait for that and you will be able to easier support this.

PR #2430, opened 2020-11-03
New Presence: Origin

mmattbtw:
Pretty finished, but not many "features"
I would like it so that the `presenceData.details` just has the webpage Title, like what shows in the tab on your browser, but I'm not sure on how to do that yet.

Visiting the homepage:
Visiting the homepage

Browsing Games:
Browsing Games

### DIFF
--- a/websites/O/Origin/presence.ts
+++ b/websites/O/Origin/presence.ts
@@ -1,0 +1,33 @@
+const path = window.location.pathname;
+const browsingStamp = Math.floor(Date.now() / 1000),
+  presence = new Presence({
+    clientId: "773080268126683186"
+  });
+let title, timestamp;
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "origin_logo",
+    smallImageKey: "shop_icon"
+  };
+
+  if (
+    document.location.pathname == "/" ||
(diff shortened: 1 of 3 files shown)


websites/O/Origin/presence.ts
```diff
@@ -0,0 +1,33 @@
...
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "origin_logo",
+    smallImageKey: "shop_icon"
+  };
+
+  if (
+    document.location.pathname == "/" ||
+    document.location.pathname == "/usa/en-us/store"
```
ririxichan, line 16: You shouldn't only restrict it to one lang.

websites/O/Origin/presence.ts
```diff
@@ -0,0 +1,33 @@
+const path = window.location.pathname;
+const browsingStamp = Math.floor(Date.now() / 1000),
```
ririxichan, line 2: merge consts

ririxichan (changes requested):
There are more pages to support and you should support them if you want us to merge this pr :) also fix.

ririxichan:
Also we do not accept drafts, so i changed it to ready to review, if you want to have draft, close this and do it locally

PR #2459, opened 2020-11-12
Rich Presence for my University (ULL) web page and virtual classroms 

Nerea-Cassian:
image


Nerea-Cassian:
I have to do some improvements

PR #2481, opened 2020-11-16
SCR Presence

scrdevelopment:


ririxichan:
We changed our guidelines yesterday about online radios and now it means: ``Presences for online radios are only allowed if the radio has at least 100 weekly listeners and 15 concurrent.`` If you want us to merge this you need to provide us statistics they are higher or equals to these, and also you need to fix issues. So for now we are closing this pull request.

PR #2511, opened 2020-11-22
KissAsian: Fix typo

TorchedSammy:


websites/K/KissAsian/presence.ts
```diff
@@ -87,7 +87,7 @@ presence.on("UpdateData", async () => {
         }
       } else if (isNaN(duration)) {
         presenceData.startTimestamp = browsingStamp;
-        presenceData.details = "Looing at:";
+        presenceData.details = "Looikng at:";
```
ririxidev, line 90: you made typo with another typo. It should be ``Looking``

ririxidev (changes requested):
and bump version

websites/K/KissAsian/presence.ts
Bas950: introducing a new typo!

TorchedSammy: :|

ririxidev (approved):

Bas950 (approved):

PR #2527, opened 2020-11-24
Added Harmony Radio Presence

xxdjbobbyxx:
I have added Harmony Radio as a Presence and removed the Symphony files as Symphony cannot be released yet.

I will attach a image of the presence working  below

image


ririxidev:
We need stats from this radio station because as far I know it only has >10 listeners and fix codacy errors.

xxdjbobbyxx:
I never saw this and that is correcct there stats are below those requirements

ririxidev:
So sorry, but we can't accept this radio station then. There are a lot of radio stations that has >10 listeners + have same software everywhere and same presence ofc (getting everything from api). So, sorry but we can't accept this then.

xxdjbobbyxx:
Can I just ask where can I find that rule?

ririxidev:
https://github.com/PreMiD/Presences/blob/master/.github/CONTRIBUTING.md

PR #2533, opened 2020-11-26
Microsoft Teams Rich Presence

MatStef132:
Hello!
I have made rich presence for microsoft teams. It supports all (I think) pages of microsoft teams, except of course extensions and manager/teacher/admin only things.
Example screenshot of one of it: https://imgur.com/a/tDwvdEm
Screenshot of settings: https://imgur.com/a/rfuPRuE

### DIFF
--- a/websites/T/Teams/dist/metadata.json
+++ b/websites/T/Teams/dist/metadata.json
@@ -12,7 +12,7 @@
     "pl": "Bądź w kontakcie z innymi i zachowaj dobrą organizację pracy. Osiągaj więcej razem z innymi — w pracy, szkole i życiu — dzięki usłudze Microsoft Teams."
   },
   "service": "Microsoft Teams",
-  "version": "1.0",
+  "version": "1.0.0",
   "logo": "https://imgur.com/7pDg1Eu.png",
   "thumbnail": "https://imgur.com/6HDv24v.png",
   "color": "#4A51BA",
--- a/websites/T/Teams/dist/metadata.json
+++ b/websites/T/Teams/dist/metadata.json
@@ -1,0 +1,161 @@
+{
+  "$schema": "https://schemas.premid.app/metadata/1.0",
+  "author": {
+    "name": "MatStef132",
+    "id": "269787765120565248"
+  },
+  "url": "teams.microsoft.com",
+  "description": {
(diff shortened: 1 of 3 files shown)


websites/T/Teams/dist/metadata.json
```diff
@@ -0,0 +1,161 @@
...
+  },
+  "url": "teams.microsoft.com",
+  "description": {
+    "en": "Stay connected and organized. Accomplish more together across work, school, and life with Microsoft Teams.",
+    "de": "Bleiben Sie gut vernetzt und organisiert – und erreichen Sie gemeinsam mehr mit Microsoft Teams: am Arbeitsplatz, in Schule und Studium und privat.",
+    "ru": "Будьте на связи и планируйте свои действия. Решайте поставленные задачи сообща и успевайте больше на работе, в школе и дома благодаря Microsoft Teams.",
+    "pl": "Bądź w kontakcie z innymi i zachowaj dobrą organizację pracy. Osiągaj więcej razem z innymi — w pracy, szkole i życiu — dzięki usłudze Microsoft Teams."
+  },
+  "service": "Microsoft Teams",
+  "version": "1.0",
```
Bas950, line 15: 1.0.0

Bas950 (changes requested):
Also what you did with messages is meh, we are almost ready to beta release an update that adds the ability  to translate presences. So you might want to wait for that.

websites/T/Teams/dist/metadata.json
MatStef132: Also to the messages I wanted an easy way for everyone to translate or change the text of the rich presence to how they want it.

Timeraa:
> Also what you did with messages is meh, we are almost ready to beta release an update that adds the ability  to translate presences. So you might want to wait for that.

It's not meh, I'll add something like that to YouTube Music etc. too since you'll be able to change a lot of texts

Timeraa (approved):

Bas950:
> It's not meh, I'll add something like that to YouTube Music etc. too since you'll be able to change a lot of texts. In terms of translations only, yes but in terms of customization nah

I mean for YouTube music you should do it that way for the format it uses when listening to a song. But I personally wouldn't add it for general text, I would use the translations.

Bas950 (approved):

PR #2570, opened 2020-12-04
VRChat: Patch 2

KnuffelBeestje:
- Implemented privacy mode into feedback.vrchat.com
- Searches are now hidden when privacy mode is turned on
- Added support for VRChat Plus

image


Bas950 (approved):
Ziet er goed uit!

ririxidev (approved):

PR #2635, opened 2020-12-16
Bubblez.app Website

ProjectVexion:


ririxidev:
We are not allowing drafts. So I'm converting it to normal PR and saying this presence is low quality right now so we are not going to accept it.

ProjectVexion:
Understandable

PR #2683, opened 2020-12-25
xero.gg | Added "Challenges" page.

Dekirai:
Whats new: The website has added a "Challenges" page so I updated my presence to make it work with the new website update.
Proof of it working: [Video](https://streamable.com/47nr6e)

### DIFF
--- a/websites/X/Xero.gg/presence.ts
+++ b/websites/X/Xero.gg/presence.ts
@@ -22,6 +22,7 @@ presence.on("UpdateData", async () => {
       "#player-profile-header-heading > div:nth-child(2) > div > div > div.medium.normal-color-name"
     ).textContent;
     if (showProfile) {
+      document.querySelector("#main > div.mixed-content > h1");
       presenceData.details = `Player: ${player_nickname}`;
       try {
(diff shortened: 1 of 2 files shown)


websites/X/Xero.gg/presence.ts
```diff
@@ -96,6 +96,17 @@ presence.on("UpdateData", async () => {
...
+    presenceData.details = "Viewing a page:";
+    presenceData.state = "Challenges";
+    try {
+      const challenges_title = document.querySelector(
+        "#settings-data-container > div.xero-header-standalone"
+      ).textContent;
+      presenceData.state += ` (${challenges_title})`;
+    } catch {
+      //Catch nothing
+    }
```
ririxidev, line 109: If you use try function, why you don't catch anything? Just use something like this:
```javascript
const f = hh() ? jj() : ""
```

ririxidev (changes requested):

Dekirai:
I just removed the trycatch since it was useless at this point anyway. Don't even know why I've added it.
Also made small changes to other trycatches where I catched nothing.

ririxidev (approved):

Bas950 (approved):

Issue #2669, opened 2020-12-22
Discord Merch | discordmerch.com


MaXfAa:
**Prerequisites and essential questions** 
- [x] Is it a popular site?
- [x] Is the website older than 2 months? 
- [ ] Is the site locked to a specific country/region?
- [ ] Is the site a paid service? (e.g. Netflix, Hulu)
- [ ] Does the website feature NSFW content? (e.g. porn, etc...)
- [ ] Are you a donator/patron?
- [x] Do you acknowledge that coding presences is completely voluntary and may take time for your service to be added regardless of priority?

**What's your Discord username?** 
MaXfAa#1068

**What's the name of the service?** 
https://discordmerch.com/ | Discord Merch (the official clothes store of Discord)

**What should the Presence display?** 
1. Discord Merch
2. The collection
3. The product

**If possible, please provide a logo for the service (512x512 minimum)** 
https://cdn.shopify.com/s/files/1/1033/1885/files/logo-transparent_185x.png?v=1603867104

ScypterDev:
I'll be able to do it!

MaXfAa:
> I'll be able to do it!

That's great ! 😀

MaXfAa:
It's over ! Perfect !

PR #2688, opened 2020-12-26
Added https://discordmerch.com/

ScypterDev:
image

Added from #2669


Bas950:
Just add it to the Discord presence

ScypterDev:
> Just add it to the Discord presence

What do you mean?

ScypterDev:
> Just add it to the Discord presence

It is a different domain is that possible?

Bas950:
> > Just add it to the Discord presence
> 
> It is a different domain is that possible?

Array of URL, or use regex.

PR #2693, opened 2020-12-27
Add presence for Exaroton

AyamDobhal:
Added a presence for [Exaroton](https://exaroton.com) - the paid service of [Aternos](https://aternos.org).

Working proof:
Screenshot 2020-12-27 160318
Screenshot 2020-12-27 160339
Screenshot 2020-12-27 160407
Screenshot 2020-12-27 160432

### DIFF
--- a/websites/E/Exaroton/presence.ts
+++ b/websites/E/Exaroton/presence.ts
@@ -64,6 +64,8 @@ presence.on("UpdateData", async () => {
           presenceData.details = "Help Center";
         }
         break;
+      default:
+        presenceData.details = "Browsing...";
     }
   }
   presence.setActivity(presenceData);
--- a/websites/E/Exaroton/presence.ts
+++ b/websites/E/Exaroton/presence.ts
@@ -66,6 +66,7 @@ presence.on("UpdateData", async () => {
         break;
       default:
         presenceData.details = "Browsing...";
+        break;
     }
   }
   presence.setActivity(presenceData);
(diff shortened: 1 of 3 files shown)


Bas950 (dismissed):

websites/E/Exaroton/presence.ts
```diff
@@ -0,0 +1,70 @@
...
+    const matchedPath = path.match(/:|software|players|addons/);
+    if (matchedPath && !matchedPath.index)
+      path = matchedPath[0] as keyof typeof paths;
+    const page = paths[path];
+    if (page) presenceData.details = page;
+  } else {
+    const page = document.location.hostname.split(".")[0];
+    presenceData.startTimestamp = Date.now();
+    switch (page) {
+      case "support":
```
ririxidev, line 40: maybe do
```default:``` ? because like you only switch and check if it is support so maybe if page will not be found, just say idk, watching support or something 🤔

ririxidev (changes requested):

websites/E/Exaroton/presence.ts
AyamDobhal: Well there isn't any possibility of this happening with Exaroton (or I might be unaware), but I'll just add a `default:` case so that in such a case, it will display **"Browsing..."** under the presence details.

ririxidev: ye ye, that's what i wanted thanks :D

ririxidev: also add ``break;`` bc u forgot

AyamDobhal: oh yes sorry

ririxidev (approved):

Bas950 (approved):

PR #2737, opened 2021-01-02
feat(websites): Add otakudesu.tv presence

KurokuTetsuya:


KurokuTetsuya:
Pending since iframe doesnt work:
image

ririxidev:
don't do it as draft then and just wait xd, because idk if it is our issue or browsers issue exactly

PR #2769, opened 2021-01-07
add HackerRank 

NMW03:
This PR adds [HackerRank ](https://www.hackerrank.com/ )presence
Screenshots:
Main
Dashboard
Challenges
Challenge
Skills Verification

### DIFF
--- a/websites/H/HackerRank/presence.ts
+++ b/websites/H/HackerRank/presence.ts
@@ -6,9 +6,9 @@ presence.on("UpdateData", () => {
   const presenceData: PresenceData = {
     largeImageKey: "logo",
     startTimestamp: Math.floor(Date.now() / 1000)
-  };
-  const path = (text: string) => { return document.location.pathname.includes(text) },
-   title = document.querySelector("h1").textContent;
+  },
+    path = (text: string) => { return document.location.pathname.includes(text) },
+    title = document.querySelector("h1").textContent;
 
   if (path("/dashboard")) {
     presenceData.details = "In dashboard";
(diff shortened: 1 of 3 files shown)


websites/H/HackerRank/presence.ts
```diff
@@ -0,0 +1,51 @@
+const presence = new Presence({
+  clientId: "655534149871992845"
+});
+
+presence.on("UpdateData", () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "logo",
+    startTimestamp: Math.floor(Date.now() / 1000)
+  },
```
Bas950, line 9: you should probs set your Date.now outside of the updatadata event

Bas950 (approved):
but the rest looks good

websites/H/HackerRank/presence.ts
NMW03: why 🤔

Bas950: Because you will set a new time now every updateData event, instead of just keep setting the same.

ririxidev (approved):

websites/H/HackerRank/presence.ts
```diff
@@ -0,0 +1,52 @@
+const presence = new Presence({
+  clientId: "655534149871992845"
+}),
+  startTimestamp = Math.floor(Date.now() / 1000);
+
+presence.on("UpdateData", () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "logo",
+    startTimestamp
```
Bas950, line 9: yea that wont work lmao you would need
`startTimestamp: startTimestamp`

Bas950 (changes requested):
but the rest looks good

Bas950 (approved):

PR #2790, opened 2021-01-12
Add official animepahe proxy domains

MulverineX:
Playback tracking is busted but there's no way to debug since temp presences can't access iframe. But this needed to get added


ririxidev (approved):

ririxidev:
Soon we will release update with fixed iframes 🤷 because maybe browser just block us etc. etc. we just need to check :p

Bas950 (approved):

PR #2807, opened 2021-01-15
Add Yandex.Translate

StrikerFRFX:
**Proof of Working:**
Discord_vxwcNZUs4l
Discord_JkoyZK7brv
Discord_zsLqvMD9MB
Discord_5qWTrX99ih
Discord_ak8W4Ez0C7

Closes #2699

PR #2817, opened 2021-01-17
Fix: Google Translate Presence

### LINKED ISSUE #2806: Google Translate Presence does not display To and From language correctly.
**Describe the Bug**
The Google Translate Presence is supposed to show what language you are translating from and what language you are translating to, but instead shows "Choosing" and "Detecting" which after looking at the code, means it is not able to properly detect the language

**Steps to Reproduce**
1. Launch Google Translate
2. Select your languages and input some text
3. Observe the Google Translate Presence

**Expected Behavior**
Should show (for example)
To: English
From: French

Instead shows 
To: Choosing
From: Detecting

**Screenshots**
image
This is with the option "Show translation type" enabled in the actual presence's settings

image
Time elapsed works fine and this is with the above option disabled

**Environment**
- OS: Windows 10
- Browser: Chrome
- Version: 87.0.4280.141 (Official Build) (64-bit) (cohort: Stable)

**Additional context**
I do know how to fix this, would it be okay if I reference this issue in a later PR with my proposed fix?

StrikerFRFX:
**Proof of Working**:
Discord_0L5CkSkx6x
Discord_4oaMyrrSdL
The above is **working** (with the setting "Show translation type" toggled in the settings).

**and Previously Broken Screenshot**:
Discord_TBjvZdxAet
The above is **broken** (NOTE: this shows when the webpage is just a "translate.google.com/" URL meaning no languages have been selected yet. It does not mean the presence is broken if this is displayed. In the broken version, this shows even with a language selected for translation (source and destination) ). "Show translation type" is not toggled in the settings in this screenshot.

However:
The same would show on the below screenshot, which is with the setting "Show translation type" toggled in the settings. The difference is this one updates upon typing in the appropriate translation fields. This one is an example of the **working** one, with just the "translate.google.com/" (no translation searches)
Discord_rVDXWr6S6G

NOTE:
The miscellaneous commits on this PR (such as 12c2929) mean that this PR MUST be merged before #2807 else breaking changes (TLDR; I had to remove all commits of Yandex.Translate in order to open this PR otherwise there would be breaking changes (2 presences in one PR) )

Resolves #2806


ririxidev (approved):

websites/G/Google Translate/dist/metadata.json
```diff
@@ -7,6 +7,11 @@
     {
       "name": "Callum",
       "id": "506899274748133376"
+    },
+
+    {
```
Bas950, line 12: remove that space

Bas950 (changes requested):

websites/G/Google Translate/dist/metadata.json
StrikerFRFX: sure can do

ririxidev (approved):

Bas950 (approved):

PR #2822, opened 2021-01-18
Removal of DevTR presence

eren885:


ririxidev:
why you want to remove it?

PR #2849, opened 2021-01-22
Exaroton -> exaroton

BluewyDev:
exaroton is supposed to not be capitalized according to its own branding rules.

### DIFF
--- a/websites/E/Exaroton/dist/metadata.json
+++ b/websites/E/Exaroton/dist/metadata.json
@@ -12,7 +12,7 @@
     "support.exaroton.com"
   ],
   "version": "1.0.0",
-  "logo": "https://i.imgur.com/uexkJD4.png",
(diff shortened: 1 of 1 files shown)


ririxidev:
ehhhh, doesn't really matter 🤔

BluewyDev:
If that doesn't matter, then i don't see why it would matter to merge it..
The way it's done in this presence goes against exaroton's guidelines, imo neglecting those entirely because to you it doesn't matter seems a bit unlogical.

ririxidev:
How big first letter is against exaroton's guidelines??? Guidelines are being accepted by user when they have account in their service, but not when you write their name somehere.

BluewyDev:
They're not public, but the employees of exaroton have mentioned this lots of times.

Thanks anyway

PR #2875, opened 2021-01-26
Add Keyma.sh

CameronCT:
Due to last pull request the version has been changed to 1.0.0 and the service name matches the folder contents and here is the screenshot of viewing a profile page.

Capture

### DIFF
--- a/websites/K/Keymash/presence.ts
+++ b/websites/K/Keymash/presence.ts
@@ -1,0 +1,58 @@
+const presence = new Presence({
+  clientId: "719415069460529163"
+});
+
+presence.on("UpdateData", () => {
+  const
+    presenceData: PresenceData = {
+      details: "Here",
+      largeImageKey: "logo"
+    },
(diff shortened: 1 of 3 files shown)


Bas950 (dismissed):

websites/K/Keymash/presence.ts
```diff
@@ -0,0 +1,58 @@
+const presence = new Presence({
+  clientId: "719415069460529163"
+});
+
+presence.on("UpdateData", () => {
+  const
+    presenceData: PresenceData = {
+      details: "Here",
```
ririxidev, line 8: for what reason you set details?

ririxidev (changes requested):

websites/K/Keymash/presence.ts
CameronCT: I pushed a change to resolve that.

ririxidev (approved):

Bas950 (approved):

PR #2876, opened 2021-01-26
Updated of kanishiel.com and gunivers.net

fantomitechno:
I know that these have just been accepted but I added a few thing : small images for the precence and privacy option 

**Proof that the changes work :**

*For gunivers :*
I added this little image when you are on the workspace
image

I also forget to add the presence when your looking for a user on the main website
image

I added a privacy mode that remove all the presence and set your status the same everywhere
image

*For kanishiel :* 
I added images for the dashboard 
image
And for the werewolves roles 
image

I added two type of privacy : one which hide the server you're editing 
image
And an other that remove all the presence and set your status the same everywhere
image

Oh and for gunivers, I have registered the url status.gunivers.net but after I speak about this presence with the website owner, he move it to another website so I removed it.

### DIFF
--- a/websites/G/Gunivers/dist/metadata.json
+++ b/websites/G/Gunivers/dist/metadata.json
@@ -6,7 +6,6 @@
     },
     "url": [
       "gunivers.net",
-      "status.gunivers.net",
       "project.gunivers.net"
     ],
     "description": {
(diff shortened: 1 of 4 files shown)


Bas950 (changes requested):
bump version

fantomitechno:
oh i forgot for gunivers

Bas950 (approved):
bump version

ririxidev (approved):

PR #2877, opened 2021-01-26
Update metada.json

JetPlet:


ririxidev:
huh? it doesn't break anything lel

PR #2912, opened 2021-01-30
youtube music: check if song is from channel

Hades785:
Add a check to `getAuthorString()` to have the year displayed only on channels, and not on user library uploads.
Prevents library uploads having the last four characters of the title be repeated.

### DIFF
--- a/websites/Y/YouTube Music/dist/metadata.json
+++ b/websites/Y/YouTube Music/dist/metadata.json
@@ -16,7 +16,7 @@
     "zh_TW": "全新的音樂服務，包含適用於Android，iOS和PC的官方專輯，單曲，視頻，混音，現場表演等。都在這裡！"
   },
   "url": "music.youtube.com",
-  "version": "2.5.1",
+  "version": "2.5.2",
   "logo": "https://i.imgur.com/cIZnPDy.png",
   "thumbnail": "https://i.imgur.com/iQ8MA50.png",
   "color": "#E40813",
--- a/websites/Y/YouTube Music/presence.ts
+++ b/websites/Y/YouTube Music/presence.ts
@@ -15,22 +15,35 @@ function getAuthorString(): string {
 
   //* Author tags more than one => YouTube Music Song listing with release year etc.
   if (authors.length > 1) {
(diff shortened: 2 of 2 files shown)


ririxidev:
Not useful, because on non user library songs can have this issue too.

Hades785:
> Not useful, because on non user library songs can have this issue too. _Tested by myself_

@ririxidev would you mind giving me one or more of such cases?
Might as well try and see if I can take care of these ones as well.

ririxidev:
When I will have some example i will give you

keanuplayz:
If I may; What's the point of immediately closing a PR, stating you have tested certain edge-cases, but not being able to demonstrate these cases?
This issue seems to be rather useful to fix, as it just looks broken if it occurs.
I recommend reconsidering, and perhaps expanding upon the changes made in this PR. I'm sure tons of people have this issue.

ririxidev:
Why? This fix would only work on songs that doesn't have channel (aka self-added and some that youtube can't assign channel). I'm able to provide cases but I don't remember any song that had this issue. When I will get them, I will provide them 😉

keanuplayz:
> This fix would only work on songs that doesn't have channel
That's precisely the point. On files that were uploaded by the user, this presence *still* fetches the last four characters and adds them as a date. This should not happen.
This PR disables the functionality of adding the date on user-uploaded files.

ririxidev:
Date should be added on user-uploaded files if it is possible... Okay so I found example. Two authors, song, no year provided: 
image
image
*polish language, but you know polubień - likes etc.) it shouldn't that it will display too. This pr wouldn't work here because these authors has channels and normally you can get channel div with provided one in pr.

Hades785:
That example is a video.
So long as there is no songs with irregularly formatted title strings (haven't seen any so far), checking if it's a video or a song should fix these cases.
I have a potential check for now (checking for the attribute `video-mode_` on the page: `document.querySelector("ytmusic-player-page[video-mode_]")`)

I am going to add the check for now.

Bas950 (approved):

ririxidev (approved):
:)

PR #2932, opened 2021-02-03
➕ Add StatifyBot Presence

tayron1:
img
img


tayron1:
_Redi is the Owner of Statify thats why he is listed as the Author_

ririxidev:
Domain is not 2 months old.

PR #2939, opened 2021-02-04
Python anywhere

Poloin34:
Added presence for Python anywhere website. Displaying the current directory or file and current user


ririxidev:
don't do another pull request from same branch. not it added the ph changes... make another branch for this presence...

PR #2984, opened 2021-02-11
Joyn Presence: Update

N0chteil:
Added compilation page


websites/J/Joyn/presence.ts
```diff
@@ -124,6 +125,144 @@ presence.on("UpdateData", async function () {
       presenceData.details = "Datenschutzerklärung";
     } else if (urlpath[1] == "agb") {
       presenceData.details = "AGB's";
+=======
```
ririxidev, line 128: it kinda broke it isnt it? Fix it and make new pull request

ririxidev (changes requested):

PR #2991, opened 2021-02-12
🐛 Fixed Bug in Kickassanime Presence

StrikerFRFX:
Anime's title and episode was not displaying properly, so I restructured the variables and changed how exactly it fetches the title and episode. Anime's with the character "-" in it would trip the shorter, previous system into thinking it should split there and that was the Episode. This is fixed now 😄 

Before fix:
C5qbkKrZ9n

After fix:
6uG9yYUqEo


websites/K/Kickassanime/presence.ts
```diff
@@ -101,20 +122,37 @@ presence.on("UpdateData", async () => {
...
-                1,
-                currentAnimeWatching[1].length
-              ));
+      currentAnimeTitle = document.querySelector("a.ka-url-wrapper")
+        .textContent;
+      currentAnimeEpisode = document
+        .querySelector("a.ka-url-wrapper:nth-child(1)")
+        .textContent.replace("Previous ", "")
+        .split("Episode ")[1];
+      if (isMovie == false) {
```
ririxidev, line 131: fr? ``!isMovie``

websites/K/Kickassanime/presence.ts
```diff
@@ -82,16 +82,37 @@ presence.on("UpdateData", async () => {
...
```
ririxidev, line 91: fr? !isMovie

ririxidev (changes requested):

Bas950 (approved):

ririxidev (approved):

PR #3038, opened 2021-02-18
Add Crunchy-DL

wchar-t:
Hello. Crunchy-DL is a website, free to use and free of ads, where you can watch the most famous animes, whose focus is to help the community by creating a simple, yet really functional environment

image
image
image
image
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ package-lock.json
 
 # Exception for extension recommendations
 !**/.vscode/extensions.json
+/.vs
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,3 @@ package-lock.json
 
 # Exception for extension recommendations
 !**/.vscode/extensions.json
-/.vs
--- a/websites/C/Crunchy-DL/presence.ts
+++ b/websites/C/Crunchy-DL/presence.ts
@@ -3,19 +3,16 @@ const presence = new Presence({
 });
 
 function handlePresenceData(presenceData: PresenceData) {
-  const viewingRoutes = ["/", "/perfil", "/categoria"];
-  const path = location.pathname;
-
-  const capitalize = (str: String) => {
-    // joke below. ignore it
-
-    return str.split(" ").map((e) => {
-      return e.split("").map((ee, ii) => {
-          return ii == 0 ? ee.toUpperCase() : ee.toLowerCase();
-      }).join("");
-    }).join(" ");
-
-  }
+  const viewingRoutes = ["/", "/perfil", "/categoria"],
+        path = location.pathname,
+        capitalize = (str: string) => {
+          // lols below.. ignore it
+          return str.split(" ").map((e) => {
+            return e.split("").map((ee, ii) => {
+                return ii == 0 ? ee.toUpperCase() : ee.toLowerCase();
+            }).join("");
+          }).join(" ");
+        };
 
   if(viewingRoutes.includes("/" + path.split("/")[1])) {
     presenceData.details = "Visualizando:";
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -17,4 +17,5 @@ package-lock.json
 **/.history
 
 # Exception for extension recommendations
-!**/.vscode/extensions.json
\ No newline at end of file
+!**/.vscode/extensions.json
+/.vs
```
ririxidev, line 21: Delete it to revert .gitignore to its state in repo. You shouldn't change this file

websites/C/Crunchy-DL/presence.ts
```diff
@@ -0,0 +1,94 @@
...
+  presence.setActivity(presenceData);
+}
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "large",
+    smallImageKey: "funi",
+    smallImageText: "CR-DL",
+    details: ".",
+    state: "."
```
ririxidev, line 85: Delete them...

websites/C/Crunchy-DL/presence.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  presence.setActivity(presenceData);
+}
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "large",
+    smallImageKey: "funi",
+    smallImageText: "CR-DL",
```
ririxidev, line 83: I don't think you need smallImage...

websites/C/Crunchy-DL/presence.ts
```diff
@@ -0,0 +1,94 @@
...
+    details: ".",
+    state: "."
+   };
+
+  if (presenceData.details == null) {
+    presence.setTrayTitle();
+    presence.setActivity();
+  } else {
+    handlePresenceData(presenceData);
```
ririxidev, line 92: Why you need this function?

ririxidev (changes requested):
You should look at this:

.gitignore
wchar-t: I reverted it to its original state. Sorry. That was my VS19 messing

websites/C/Crunchy-DL/presence.ts
wchar-t: Deleted

websites/C/Crunchy-DL/presence.ts
wchar-t: It's not necessary. I removed it

websites/C/Crunchy-DL/presence.ts
wchar-t: Instead of dealing with the pages and all logic inside the else block, i created the function. The function is all about changing presenceData's details and state

ririxidev: It isn't really good option but okay I guess, will see what Bas will say about it

ririxidev (approved):

websites/C/Crunchy-DL/presence.ts
```diff
@@ -0,0 +1,90 @@
...
+    presenceData.details = "Visualizando Anime " + prependFavorite;
+    presenceData.state = document.querySelector(".anime-name").textContent;
+
+  } else if(path.includes("/episodio")) {
+    
+    try {
+      const getTime = (str: string) => {
+        const parts = str.split(":").map(e => parseInt(e));
+        return parts[0] * 60 + parts[1];
+      },
```
ririxidev, line 52: also I guess, you could use our function from Presence class
image

websites/C/Crunchy-DL/presence.ts
```diff
@@ -0,0 +1,90 @@
...
+      episodeNumber = path.split("/").slice(-2, -1)[0].padStart(2, "0"),
+      mountedEpi = `COL ${collectionNumber} - EPI ${episodeNumber}`;
+
+      presenceData.endTimestamp = new Date().getTime() + (getTime(timeDuration) - getTime(timeElapsed)) * 1000;
+      
+      presenceData.details = videoName;
+      presenceData.state = mountedEpi;
+    } catch(e) {
+      void(0);
+    }
```
Bas950, line 67: why a try catch? Just handle it

websites/C/Crunchy-DL/presence.ts
Bas950: probs better to do it inside the updatedata

Bas950 (changes requested):

websites/C/Crunchy-DL/presence.ts
wchar-t: that was part of the debugging, i guess. Now it's being handled

websites/C/Crunchy-DL/presence.ts
Bas950: ^

Bas950 (changes requested):

PR #3128, opened 2021-02-26
TETR.IO presence, Saltybet settings update and Cuevana3 button fix

magumboi:
Screenshot_33
Screenshot_35
Screenshot_36

### DIFF
--- a/websites/C/Cuevana 3/presence.ts
+++ b/websites/C/Cuevana 3/presence.ts
@@ -30,14 +30,20 @@ presence.on("UpdateData", async () => {
     document.location.pathname.includes("/episodio")
   ) {
     const timestamps = presence.getTimestamps(
-        Math.floor(video.currentTime),
-        Math.floor(video.duration)
+      Math.floor(video.currentTime),
+      Math.floor(video.duration)
       ),
       titulo = document.querySelector(`h1.Title`).textContent,
       subtitulo = document.querySelector(`h2.SubTitle`).textContent;
 
-    if (!document.location.pathname.includes("/episodio"))
+    if (!document.location.pathname.includes("/episodio")) {
       presenceData.details = titulo;
+      presenceData.buttons = [
+        { label: "Ver Película",
+          url: window.location.href
+        }
+      ];
+    }
     else {
       presenceData.details = titulo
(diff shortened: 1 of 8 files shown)


magumboi:
Screenshot_41
Screenshot_40
Screenshot_39
Screenshot_38
Screenshot_37
Screenshot_42

websites/C/Cuevana 3/presence.ts
```diff
@@ -30,21 +30,32 @@ presence.on("UpdateData", async () => {
     document.location.pathname.includes("/episodio")
   ) {
     const timestamps = presence.getTimestamps(
-        Math.floor(video.currentTime),
-        Math.floor(video.duration)
+      Math.floor(video.currentTime),
+      Math.floor(video.duration)
       ),
```
Bas950, line 35: just use presence.getVideoTimestamps or how its called

Bas950 (changes requested):

websites/C/Cuevana 3/presence.ts
magumboi: K Let me check

magumboi: these are the only functions I found
image

magumboi: ok i think i got it

ririxidev: presence.getTimestampsfromMedia()

magumboi: I don't think I can use it since the iframe.ts is doing all the heavy lifting of the video, thats why I'm using presence.getTimestamps() instead. 🤔

magumboi: I don't think I can use it since the iframe.ts is doing all the heavy lifting of the video (the video is within an iframe), thats why I'm using presence.getTimestamps() instead. 🤔

magumboi: tried to do something like this:
image
but ending up getting undefined 😢:
image

Bas950 (approved):

ririxidev (approved):

PR #3183, opened 2021-03-06
Bump Jstris Ver to 1.1.2 - Emergency Hotfix to Fix Presence for Non-Signed in Users.

jy1263:
With signed in:
image

Without signed in:
image

Buttons with Sign In:
image

Buttons without Sign In:
image


Bas950 (changes requested):
bump version

StrikerFRFX (changes requested):
Show proof of presence still working for a non signed in user as well as a signed in user

jy1263:
oki

StrikerFRFX (approved):
As long as that last button on your 3rd screenshot is removed when they're not signed in, all good (if they can access that page or whatever when not signed in)

Bas950 (approved):

PR #3188, opened 2021-03-06
Update presence.ts

trains2050:
made it better

### DIFF
--- a/websites/M/MKPC/presence.ts
+++ b/websites/M/MKPC/presence.ts
@@ -15,14 +15,12 @@ presence.on("UpdateData", async () => {
     document.location.pathname == "/fr.php" ||
     document.location.pathname == "/en.php"
   ) {
-    
     presenceData.details = "Viewing home page";
   } else if (document.location.pathname == "/forum.php") {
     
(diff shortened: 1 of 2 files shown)


websites/M/MKPC/presence.ts
```diff
@@ -57,5 +51,6 @@ presence.on("UpdateData", async () => {
     presence.setActivity();
   } else {
     presence.setActivity(presenceData);
+    presenceData.startTimestamp = browsingStamp;
```
ririxidev, line 54: you are adding timestamp after setting activity? what?

ririxidev (changes requested):

websites/M/MKPC/presence.ts
Bas950: 😂

Bas950 (changes requested):

EncryptedDev:
Not really going to do much in terms of the presence, but why is there an empty line on line 20 of presence.ts? 
Just looks a bit strange having a random line in the middle of an else if statement lol

PR #3190, opened 2021-03-06
Update metadata.json

berksmbl:
Service name edited


ririxidev (changes requested):
service name must be the same as folder name.

PR #3216, opened 2021-03-08
Added Buttons

Gurrman375:
image
image

Added Buttons:
image
image
image

Note: For some reason elapsed time doesn't work with buttons. I think I just have a bad installation of PreMiD.


Bas950 (changes requested):
Enable the infos or remove the comments.

Bas950 (changes requested):
bump version

Bas950 (approved):

StrikerFRFX (approved):

PR #3221, opened 2021-03-09
🎉 Tailwind CSS and OneDrive Presences!

tarikcoskun:
A utility-first CSS framework that can be composed to build any design, directly in your markup.

moment1

moment2

moment3

moment4


tarikcoskun:
OneDrive is an online file storage and sharing service from Microsoft.

moment1

moment2

moment3

moment4

EncryptedDev (changes requested):
You shouldn't use setActivity many times in the presence, you should use if, else if, else statements to declare presence details and at the end of the file, outside of your if, else if and else statements, but in your overall UpdateData, you should use <presence>.setActivity()

tarikcoskun:
That was my first Presence so I was testing stuff and there are just some junks I left behind

tarikcoskun:
@Bas950 got everything done, i guess

StrikerFRFX (approved):

Bas950 (approved):

PR #3239, opened 2021-03-12
👽 Updated video playback parser for Jellyfin 10.7.0

alexbcberio:
There've been some changed in the frontend and the video playback has to be updated, the other features work ok.

Proof of video playback
image
image

### DIFF
--- a/websites/J/Jellyfin/presence.ts
+++ b/websites/J/Jellyfin/presence.ts
@@ -697,18 +697,12 @@ async function setDefaultsToPresence(): Promise<void> {
   if (presenceData.smallImageText) {
     delete presenceData.smallImageText;
   }
-  /*
   if (presenceData.startTimestamp) {
     delete presenceData.startTimestamp;
   }
   if (presenceData.endTimestamp && isNaN(presenceData.endTimestamp)) {
     delete presenceData.endTimestamp;
   }
-  */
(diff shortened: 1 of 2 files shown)


websites/J/Jellyfin/presence.ts
```diff
@@ -701,12 +697,18 @@ async function setDefaultsToPresence(): Promise<void> {
...
   }
-  if (presenceData.endTimestamp) {
+  if (presenceData.endTimestamp && isNaN(presenceData.endTimestamp)) {
     delete presenceData.endTimestamp;
   }
+  */
+
+  // set default values, required for Math.floor on devHelper.ts
+  presenceData.startTimestamp = 0;
+  presenceData.endTimestamp = 0;
```
Bas950, line 711: make sure to delete them if you dont set them

Bas950 (changes requested):

websites/J/Jellyfin/presence.ts
alexbcberio: I was previously deleting them, but if they are unset I get `NaN` on those fields.

I revert is to the previous code, deleting the values

Bas950 (approved):

StrikerFRFX (approved):

PR #3285, opened 2021-03-19
Added Google Photos Presence!

JohanRBX:
unknown (5)
unknown (4)

### DIFF
--- a/websites/G/Google Photos/dist/metadata.json
+++ b/websites/G/Google Photos/dist/metadata.json
@@ -1,0 +1,26 @@
+{
+    "author": {
+      "name": "Johan.",
+      "id": "480176818297503744"
+    },
+    "contributors": [
+      {
+        "name": "DiamondGun",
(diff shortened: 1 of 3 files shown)


ririxidev:
Hi @JohanRBX, we have some news for you:

We are so sorry, but your presence is in very low quality and Isn't eligible for merging. We can provide you some informations what is wrong:
 - ``var`` are not a thing for us. You should only use ``const`` or ``let``. If you don't know their difference, try checking it at google.
 - You are doing everytime same object (presenceData) but with different state/details. This shouldn't happen! Try doing one object and then assigning state/details by e.g. ``presenceData.state = "your thing"``
 - Merging variables is important thing too! You can check how it works on ESLint rules page: https://eslint.org/docs/rules/one-var

Hope it will help you and we will wait for your fixed presence in a future!

Thanks,
ririxi (Contact: mail@ririxi.dev, ririxi#2721)

PR #3393, opened 2021-03-31
✨ One singular Irish translation for the bois

callumok2004:
Few changes/fixes.

We noticed sometimes, though rarely when saving a file, the version would bump but the translation would not be added. This is a temporary solution to that.


ririxidev:
Remove the Translation Tool commit, I will add it to mine because I will push it in few minutes.

Bas950 (approved):
LGTM

PR #3407, opened 2021-04-02
Mugzone:updated some icon and timer

Himori-0:
I changed two icon name and change the timer for it to show in everypage instead of a few pages only
image
image


ririxidev:
Better update fork since it shows that you add mugzone again as whole presence.

PR #3427, opened 2021-04-04
Fixed and updated Mixcloud Presence

millsoft:
Mixcloud didn't work at all because of wrong if/else and typescript errors (eg. tried to access variables which were not available).
Also refactored the code to not use hashed class names, which were changed after a new mixcloud-deploy. I updated the presence to use more generic class names which should not change soon. Tested it with normal mixcloud music and live streams.
The added contributor MilMike is me in discord.

### DIFF
--- a/websites/M/Mixcloud/presence.ts
+++ b/websites/M/Mixcloud/presence.ts
@@ -4,7 +4,7 @@ const presence = new Presence({
   strings = presence.getStrings({
     play: "presence.playback.playing",
     pause: "presence.playback.paused",
-    live: "presence.activity.live",
+    live: "presence.activity.live"
   });
 
 let author: string, title: string, url: string;
@@ -18,14 +18,13 @@ presence.on("UpdateData", async () => {
     const normalIsPlaying: boolean =
       document
         .querySelector("div[class^='PlayButton__PlayerControl']")
-        ?.getAttribute("aria-label") === "Pause";
-
-    const liveIsPlaying: boolean =
+        ?.getAttribute("aria-label") === "Pause"
+    ,
+    liveIsPlaying: boolean =
       document
         .querySelector("[class^=LiveVideo__VideoContainer] .shaka-play-button")
-        ?.getAttribute("icon") === "pause";
-
-    const isPlaying = normalIsPlaying || liveIsPlaying;
+        ?.getAttribute("icon") === "pause",
+     isPlaying = normalIsPlaying || liveIsPlaying;
 
     if (normalIsPlaying) {
       const normalDetails = document.querySelector(
--- a/websites/M/Mixcloud/presence.ts
+++ b/websites/M/Mixcloud/presence.ts
@@ -4,72 +4,72 @@ const presence = new Presence({
   strings = presence.getStrings({
     play: "presence.playback.playing",
     pause: "presence.playback.paused",
-    live: "presence.activity.live"
+    live: "presence.activity.live",
   });
 
-let live, prevLive, author: string, title: string;
+let author: string, title: string, url: string;
 
 presence.on("UpdateData", async () => {
   const player = document.querySelector(
-    ".PlayerControls__PlayerContainer-vo7mt3-4"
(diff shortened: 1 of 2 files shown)


websites/M/Mixcloud/presence.ts
```diff
@@ -4,72 +4,72 @@ const presence = new Presence({
...
 
     const data: PresenceData = {
       details: title,
       state: author,
       largeImageKey: "mixcloud",
-      smallImageKey: paused ? "pause" : "play",
-      smallImageText: paused ? (await strings).pause : (await strings).play
+      smallImageKey: isPlaying ? "play" : "pause",
+      smallImageText: isPlaying ? (await strings).play : (await strings).pause,
+      buttons: [{ label: 'OPEN', url: url }]
```
ririxidev, line 57: what is this button for bc i dont get it?
image

websites/M/Mixcloud/presence.ts
```diff
@@ -4,72 +4,72 @@ const presence = new Presence({
   strings = presence.getStrings({
     play: "presence.playback.playing",
     pause: "presence.playback.paused",
-    live: "presence.activity.live"
+    live: "presence.activity.live",
```
ririxidev, line 7: delete this ``,`` lol

websites/M/Mixcloud/presence.ts
```diff
@@ -4,72 +4,72 @@ const presence = new Presence({
...
       data.smallImageText = (await strings).live;
     }
 
-    if (paused) {
-      delete data.startTimestamp;
-      delete data.endTimestamp;
+    if (isPlaying) {
+      presence.setActivity(
+        data,
+        isPlaying
```
ririxidev, line 68: you don't need to pass second argument.

ririxidev (changes requested):
Heya @millsoft,
You should check out thing(s) that I requested to change in my review. Please look at them!
- Also fix codacy ty

Thanks,
**ririxi**. PreMiD Reviewer Team.
Contact: mail@ririxi.dev, ririxi#2721

websites/M/Mixcloud/presence.ts
millsoft: the button should open the current music the user is listening. Either the normal played music or the live stream.

millsoft: I can remve the button, it seems not to work anyways.. No button works for me, for example the youtube presence, shows 2 buttons but nothing happens if I click it, no idea why. Maybe this is just me and it works for everyone else. (using the windows discord app client)

ririxidev: Yeah, it's discord thing. They work but they don't work for yourself. I mean, okay then? Could you show me example where it will redirect? by logging or so?

millsoft: I added a console.log when it is playing for the target url, hope a video is ok? if not I can add console.log and push and you can try by yourself (I think Codacy / ESlint might have a problem with console.log then because of "no-console")

https://ttprivatenew.s3.amazonaws.com/pulse/pits-michel/attachments/16444882/TinyTake04-04-2021-02-12-22.mp4

so the url is basically the url of the played stream. If a user opens the urk, he will land on the stream page where the stream can be played / commented etc.. or in the case of a live stream the stream url will be opened with chat etc.

ririxidev: Okay cool, could you change text to like idk `View Song` or something? And will be good to accept :D

millsoft: I just pushed a change. The button will change based on what is played. If a user is watching a live stream the button label is "View Livestream". And if the user is only listening to normal music it will show "Listen to Show". Mixcloud calls these non live recorded streams "Shows" not just song. It only allows longer music streams / mixes. So I named the button also "Shows". Hope it is ok. Let me know if there is still stomething I should do.

ririxidev (approved):
lgtm

StrikerFRFX (approved):
lgtm 2

PR #3464, opened 2021-04-08
Update multiple presences (Gamepedia domain merge)

Hans5958:
<h3>Reviewers, please read!</h3>

## Description

Well, this is awkward. 

In a recent turn of event, [the people on Fandom decided that it would be a good idea to migrate Gamepedia wikis to the Fandom domain](https://community.fandom.com/wiki/User_blog:MisterWoodhouse/The_Future_of_Gamepedia). This means that any wikis on https://gamepedia.com (like https://minecraft.gamepedia.com) is going to reside on https://fandom.com (like https://minecraft.fandom.com). SEO improvement is their main reason, if you believe.

<details><summary>Google's Featured Snippet summary, because it somehow sums up perfectly.</summary>

Today, we have a big announcement. After nearly two years of working together to bring Fandom and Gamepedia wikis onto one platform, we will be starting to migrate all Gamepedia wikis to a Fandom.com domain in early 2021.
</details>

So, I need to do a bug fix, even if it's quite late. But, when fixing it, I need to choose between two ways to do it.

1. Let the Gamepedia presence read fandom.com and keep the Fandom presence from reading it, and vice versa.
   This is the choice that I think makes more sense for the users. Both of these services are separate, and has been separated on the store since a while ago, so it would be confusing if it got merged. Also, this going to be confusing 
   The downside is that I need to code it so the Gamepedia presence only runs on Gamepedia wikis, and the Fandom presence only runs on Fandom wikis. I have also made sure that it won't run the "event listener" part of the code (or whatever is `presence.on("UpdateData", async () => {})` called). This is untested, because nobody can load two presences at once, but theoretically it's going to be fine.
2. Deprecate Gamepedia presence in favor of the Fandom presence.
   This make sense developer-facing, but it just seems off. As I have mentioned above about the problem of merging it, it's just going to be weird if a Gamepedia wiki will appear as Fandom on Discord, because both of them are different, and I can't just change the client ID as I wish. Also, looking at the architecture of the store system, I think deprecating Gamepedia is going to be a pain in the arse.

After these consideration, I decided to take the first way. I hope it is going to be fine, and it is also fine to you, reviewers.

By the way, I have also fixed my fork-copying code, so I hope it doesn't get bumped the next time `SE` gets executed.

This PR includes updates to the following presences:

- Fandom (1.2.2)
- Gamepedia (1.2.2)

*Do these need a minor bump (1.3.0) or a patch bump (1.2.2)? Please tell me in the comments below.*

## Changelog

- Update base script
  - Only start updating when there is presence data
- Do not run Fandom presence code on Gamepedia wikis... (Fandom)
- ...and vice versa (Gamepedia)
- Remove `presence.info()` debugging lines (Gamepedia)
- Adjust RegEx and code to accommodate domain change (Gamepedia)
- Sync code from Fandom (Gamepedia)

## Images

| Image | Link visited |
| ----- | ------------ |
|  | https://www.fandom.com/ |
|  | https://cavestory.fandom.com/wiki/Quote |
|  | https://www.gamepedia.com/ |
|  | https://minecraft.fandom.com/wiki/Minecraft_Plus%21 |

### DIFF
--- a/websites/F/Fandom/dist/metadata.json
+++ b/websites/F/Fandom/dist/metadata.json
@@ -6,7 +6,7 @@
   },
   "service": "Fandom",
   "url": "fandom.com",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "logo": "https://i.imgur.com/sko4P3c.png",
   "thumbnail": "https://i.imgur.com/VrVSD2b.png",
   "color": "#002A32",
--- a/websites/F/Fandom/presence.ts
+++ b/websites/F/Fandom/presence.ts
@@ -125,6 +125,12 @@ const updateCallback = {
 		
 		*/
 
+    if (!document.querySelector("#globalNavigation")) {
+      // Do not run on Gamepedia wikis.
+      presenceData = null;
+      return;
+    }
+
     let title: string, sitename: string;
     const actionResult = (): string =>
         getURLParam("action") || getURLParam("veaction"),
@@ -140,7 +146,7 @@ const updateCallback = {
       };
 
     try {
-      title = document.querySelector(".page-header__title").innerHTML;
+      title = document.querySelector("h1").textContent;
     } catch (e) {
       title = titleFromURL();
     }
@@ -284,7 +290,6 @@ const updateCallback = {
       }
     }
 
-    presenceData.startTimestamp = browsingStamp;
     if (presenceData.state) presenceData.state += " | " + sitename;
     else presenceData.state = sitename;
   } else if (currentPath[0] === "f") {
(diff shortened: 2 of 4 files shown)


Bas950 (approved):
LGTM

_but once they have fully moved to fandom the Gamepedia presence should probs go_

StrikerFRFX (approved):
LGMT, agreed with Bas above ^

Hans5958:
> but once they have fully moved to fandom the Gamepedia presence should probs go

Well, we have no clear information for that, so we will see in the future.

> ### What is going to happen to Gamepedia as a brand and a domain?
> 
> This is the `#1` question we anticipated this announcement would prompt. We have not yet reached an answer on this internally, but we felt that disclosing the Unified Community Platform plan now was best for the community, as it allows us to have a more thorough two-way dialogue about changes and enhancements. We’re not rushing to an answer because a right answer is better than a quick answer. To be clear, the decision to unify the platforms is separate from any branding decision for Gamepedia.
> 
> When we have an answer on this, it will be announced. 

—[Will "MisterWoodhouse" Kavanagh, Global Communications Lead for Fandom and Gamepedia](https://community.fandom.com/wiki/User_blog:MisterWoodhouse/Our_first_update_on_the_new_platform#What_is_going_to_happen_to_Gamepedia_as_a_brand_and_a_domain)

By the way, the next update will probably about [the new design for both Fandom and Gamepedia](https://community.fandom.com/wiki/User_blog:MisterWoodhouse/Introducing_FandomDesktop,_the_new_look_and_feel_for_desktop_users).

PR #3468, opened 2021-04-08
Added DogeHouse.tv Presence

JackCrispy:
Added DogeHouse.tv Presence

Proof it works: https://cdn.discordapp.com/attachments/820641122669428796/829752754703695892/unknown.png


ririxidev:
Domain not 2 months old.

PR #3511, opened 2021-04-12
Update multiple presences (Gamepedia domain merge 2)

Hans5958:
## Description

Well, it looks that my first attempt on resolving this issue kind of failed. Here's the second one, along with additional additions that are astonishingly overlooked.

By the way, I need to wrap the entire code in an if statement, so that's why the changes looked a lot. [This is the commit that I made on my side](https://github.com/Hans5958/PreMiD-Presences-Personal/commit/838a89d01adebc110479add5d3edf94e548b470b).

This PR includes updates to the following presences:

- Fandom (1.2.3)
- Gamepedia (1.2.3)

## Changelog

- Change method of early termination to fix critical bug where Fandom and Gamepedia presence can't run at the same time
- Add language code on the state/details for other languages than English (feature from WP)
- Revert feature "updating when there is presence data"
- Fix regex to only allow select Gamepedia domain (Gamepedia)
- Use `decodeURIComponent()` instead of `decodeURI()` for more correct title from the URL
- Use "!==" instead of negated "===" (DeepScan) (Gamepedia)
- Change position to save `let`/`const` keyword (@Bas950)

## Images

| Image | Link visited |
| ----- | ------------ |
|  | https://www.fandom.com/ |
|  | <https://nier.fandom.com/wiki/NieR_Replicant_ver.1.22474487139...> |
|  | https://www.gamepedia.com/ |
|  | https://minecraft.fandom.com/wiki/Java_Edition_1.16.5 |

### DIFF
--- a/websites/F/Fandom/presence.ts
+++ b/websites/F/Fandom/presence.ts
@@ -1,356 +1,370 @@
-const presence = new Presence({
-  clientId: "644400074008297512"
-});
+if (
+  !(
+    (
+      document.location.pathname.includes("/wiki/") &&
+      !document.querySelector("#globalNavigation")
+    )
+    // Only run on Fandom wikis.
+  )
+)
+  ((): void => {
+    const presence = new Presence({
+        clientId: "644400074008297512"
+      }),
+      browsingStamp = Math.floor(Date.now() / 1000);
+    let currentURL = new URL(document.location.href),
+      currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/"),
+      presenceData: PresenceData = {
+        details: "Viewing an unsupported page",
+        largeImageKey: "lg",
+        startTimestamp: browsingStamp
+      };
+    const updateCallback = {
+        _function: null as () => void,
+        get function(): () => void {
+          return this._function;
+        },
+        set function(parameter) {
+          this._function = parameter;
+        },
+        get present(): boolean {
+          return this._function !== null;
+        }
+      },
+      /**
+       * Initialize/reset presenceData.
+       */
+      resetData = (
+        defaultData: PresenceData = {
+          details: "Viewing an unsupported page",
+          largeImageKey: "lg",
+          startTimestamp: browsingStamp
+        }
+      ): void => {
+        currentURL = new URL(document.location.href);
+        currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/");
+        presenceData = { ...defaultData };
+      },
+      /**
+       * Search for URL parameters.
+       * @param urlParam The parameter that you want to know about the value.
+       */
+      getURLParam = (urlParam: string): string => {
+        return currentURL.searchParams.get(urlParam);
+      };
 
-let currentURL = new URL(document.location.href),
-  currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/");
-const browsingStamp = Math.floor(Date.now() / 1000);
-let presenceData: PresenceData = {
-  details: "Viewing an unsupported page",
-  largeImageKey: "lg",
-  startTimestamp: browsingStamp
-};
-const updateCallback = {
-    _function: null as () => void,
-    get function(): () => void {
-      return this._function;
-    },
-    set function(parameter) {
-      this._function = parameter;
-    },
-    get present(): boolean {
-      return this._function !== null;
-    }
-  },
-  /**
-   * Initialize/reset presenceData.
-   */
-  resetData = (
-    defaultData: PresenceData = {
-      details: "Viewing an unsupported page",
-      largeImageKey: "lg",
-      startTimestamp: browsingStamp
-    }
-  ): void => {
-    currentURL = new URL(document.location.href);
-    currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/");
-    presenceData = { ...defaultData };
-  },
-  /**
-   * Search for URL parameters.
-   * @param urlParam The parameter that you want to know about the value.
-   */
-  getURLParam = (urlParam: string): string => {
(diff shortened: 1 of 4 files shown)


websites/F/Fandom/presence.ts
```diff
@@ -1,356 +1,373 @@
...
+                presenceData.startTimestamp = timestamps[0];
+                presenceData.endTimestamp = timestamps[1];
+              } else {
+                delete presenceData.startTimestamp;
+                delete presenceData.endTimestamp;
+              }
+            } catch (e) {
+              delete presenceData.startTimestamp;
+              delete presenceData.endTimestamp;
+            }
```
ririxidev, line 109:
```suggestion
                presenceData.endTimestamp = timestamps[1];
              } else
                delete presenceData.endTimestamp;
            } catch (e)
              delete presenceData.endTimestamp;
```
i think I did it good

ririxidev (changes requested):

websites/F/Fandom/presence.ts
Hans5958: Not sure if that's okay. I'm following [the docs](https://docs.premid.app/en/dev/presence/class#gettimestampsnumber-number).

ririxidev: idk who did it but if you have endTimestamp you don't need startTimestamp.

ririxidev: image
basically, if you have endTimestamp it shows remaining no matter what

websites/G/Gamepedia/presence.ts
```diff
@@ -1,266 +1,281 @@
...
+if (
+  !(
+    (
+      document.location.hostname.endsWith("fandom.com") &&
+      document.location.pathname.includes("/wiki/") &&
+      !document.querySelector("#netbar")
+    )
+    // Only run on Gamepedia wikis.
+  ) &&
+  !(document.location.hostname === "www.fandom.com")
```
ririxidev, line 10: image
also fix deepscan yeah didn't notice it.

ririxidev (changes requested):

websites/F/Fandom/presence.ts
Hans5958: But, here's a thing. I'm not sure if there are hidden intricacies related to the RPC. I'm not sure if Discord actually uses only either one of them, but I just wanted to play safe like what I've done since the inception of this presence.

I would like to keep the `startTimestamp` as this seems the standard implementation in a lot of presences, [including YouTube](https://github.com/PreMiD/Presences/blob/master/websites/Y/YouTube/presence.ts#L240-L241) and this seemed next to unnecessary, but if is really, really a necessary change, please let me know. I would still fix the next error.

ririxidev: Looks it works like that:
 - If ``startTimestamp`` is in RPC object: Discord shows elapsed.
 - If ``endTimestamp`` is in RPC object: Discord shows remaining.
 - If ``startTimestamp`` and ``endTimestamp`` is in RPC object: Discord shows remaining. 
 That's how it works. I know that other presences use it, but literally I'm not going to tell creator aka Bas950 to delete it.

ririxidev: waiting...

Hans5958: Hello, it's been a while since I touched GH for a while. I was both shifted to my other smaller projects, and preparing for the college entry exams. Kind of surprised that this doesn't get merged after a while.

I forgot to tell you, but I have asked Bas950 about this, and he told me to "keep it like the rest of the presences." This is conflicting, so I hope you can resolve it with him. But for now, I will leave this as-is. 

I just hope that this could be merged in a quick fashion due to a critical bug that I fixed a while ago, which kind of overshadowed by this issue.

websites/F/Fandom/presence.ts
```diff
@@ -1,356 +1,373 @@
...
-  },
-  /**
-   * Initialize/reset presenceData.
-   */
-  resetData = (
-    defaultData: PresenceData = {
+    let currentURL = new URL(document.location.href),
+      currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/");
+    const browsingStamp = Math.floor(Date.now() / 1000);
+    let presenceData: PresenceData = {
```
Bas950, line 18:
```suggestion
    const browsingStamp = Math.floor(Date.now() / 1000);
    let currentURL = new URL(document.location.href),
      currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/"),
      presenceData: PresenceData = {
```

Bas950 (changes requested):

websites/F/Fandom/presence.ts
ririxidev: Bas agreed with me rn after some discussion so you are feel free to delete this line.

websites/F/Fandom/presence.ts
Hans5958: Yep. That's going to be an base update. Expect updates for all 19 prences.

websites/F/Fandom/presence.ts
Hans5958: Alright, thank god we got that sorted out on the back. I have did the changes that you requested, but unfortunately I need to add a single line since it seems that you assume `startTimestamp` is empty.

Two things that I wanted to tell you is that (a) if something goes wrong, you will be at the fault, and (b) I hope this kind of debate won't be repeated again in the future.

In the end, I apologize for the inconvenience.

ririxidev: Everything is ok, just test if that works on this specific subpage.

ririxidev (approved):

Bas950 (approved):

PR #3523, opened 2021-04-13
Bionluk | Bionluk.com

ReXulEc:
Most freelancer in Turkey and globally using this site. And most of those who use this site use apps like Discord etc. . That's why I wanted to make this presence.

Presence's photo : https://i.imgur.com/2xGsNBs.png

Best regards
-Mert Doğu / rexulec.com

### DIFF
--- a/websites/B/Bionluk/dist/metadata.json
+++ b/websites/B/Bionluk/dist/metadata.json
@@ -1,0 +1,22 @@
+{
+  "author": {
+    "name": "ReXulEc",
+    "id": "709376748206293002"
+  },
+  "service": "Bionluk",
+  "description": {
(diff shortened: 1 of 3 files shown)


ReXulEc:
I wrote var instead of const, can you please change it? Thanks!

SlowLife1661:
> I wrote var instead of const, can you please change it? Thanks!
No one use var nowadays unless you are old people kekw, but no please follow the standard

ReXulEc:
This is not low quality, there are many Turkish design servers and a lot of people need this presence.

ririxidev:
Presence is definetely still low quality. ``var``s, everywhere you create new object, no DOM readings.

PR #3580, opened 2021-04-25
[KissAsian] fix and small refactor

mlunax:
Resolve #3566

PR #3588, opened 2021-04-26
Added new url domain .li

Nitineshwar:
Kissasian domain now redirects to .li instead of .sh so relevant changes made to presence.


ririxidev:
Duplicate of #3580

PR #3607, opened 2021-04-28
Create TraplineFM Presence

PaxlPlays:


ririxidev:
that doesn't work like that unfournetely. https://docs.premid.app/dev/presence

PR #3667, opened 2021-05-05
Added Manga To Crunchyroll & Added Watch Episode Buttons

ScarVite:
POW:

Discord_hSm2N0779a
Discord_AudmUOOJSI
Discord_lbdBV0nMrm

### DIFF
--- a/websites/C/Crunchyroll/presence.ts
+++ b/websites/C/Crunchyroll/presence.ts
@@ -26,7 +26,7 @@ function getTimestamps(
 }
 
 var lastPlaybackState = null;
-var playback: Boolean;
+var playback: boolean;
 var browsingStamp = Math.floor(Date.now() / 1000);
 
 if (lastPlaybackState != playback) {
@@ -53,15 +53,15 @@ presence.on("UpdateData", async () => {
   };
 
   if (!playback && document.location.pathname.includes("/manga")) {
-    var reading: boolean = false;
+    let reading: boolean = false;
 
     if(document.location.pathname.includes("/read")) {
       browsingStamp = Math.floor(Date.now() / 1000);
-      var title = document.querySelector(".chapter-header a").innerHTML;
-      var currChapter = document.querySelector(".chapter-header").innerHTML.split("</a>")[1].split("\n")[0];
-      var currPage = document.querySelector(".first-page-number").innerHTML;
+      let title = document.querySelector(".chapter-header a").innerHTML;
+      let currChapter = document.querySelector(".chapter-header").innerHTML.split("</a>")[1].split("\n")[0];
+      let currPage = document.querySelector(".first-page-number").innerHTML;
       currPage = currPage == "" ? "1" : currPage;
-      var lastPage = document.querySelector(".images").children.length;
+      let lastPage = document.querySelector(".images").children.length;
 
       presenceData.details = title;
       presenceData.state = `${(await strings).reading} ${currChapter}`;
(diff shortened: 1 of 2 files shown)


websites/C/Crunchyroll/presence.ts
```diff
@@ -32,7 +36,16 @@ if (lastPlaybackState != playback) {
 
 var iFrameVideo: any, currentTime: any, duration: any, paused: any;
```
ririxidev, line 37: Well no `var` nor `any` type

websites/C/Crunchyroll/presence.ts
```diff
@@ -48,7 +61,52 @@ presence.on("UpdateData", async () => {
...
+      presenceData.details = (await strings).browse;
+      presenceData.startTimestamp = browsingStamp;
+
+      delete presenceData.state;
+      delete presenceData.smallImageKey;
+
+      reading = true;
+    }
+
+    presence.setActivity(presenceData, reading);
```
ririxidev, line 106: Don't need to set second argument so please remove `let reading` variable too

ririxidev (changes requested):
remove `getTimestamps` function from code with comments and use `presence.getTimestamps()` from `Presence` class
and if you are editing code already remove on line `135`, `presenceData.startTimestamp = timestamps[0]` thank you.

ririxidev (approved):
thank you 😃  lgtm

websites/C/Crunchyroll/presence.ts
```diff
@@ -48,7 +47,48 @@ presence.on("UpdateData", async () => {
...
 
-  if (!playback) {
+  if (!playback && document.location.pathname.includes("/manga")) {
+
+    if(document.location.pathname.includes("/read")) {
+      const title = document.querySelector(".chapter-header a").innerHTML,
+      currChapter = document.querySelector(".chapter-header").innerHTML.split("</a>")[1].split("\n")[0];
+      let currPage = document.querySelector(".first-page-number").innerHTML;
+      currPage = currPage == "" ? "1" : currPage;
+      const lastPage = document.querySelector(".images").children.length;
```
ririxidev, line 57:
```suggestion
      const title = document.querySelector(".chapter-header a").innerHTML,
      currChapter = document.querySelector(".chapter-header").innerHTML.split("</a>")[1].split("\n")[0],
      lastPage = document.querySelector(".images").children.length,
      currPage = document.querySelector(".first-page-number").innerHTML === "" ? "1" : currPage;
```

websites/C/Crunchyroll/presence.ts
```diff
@@ -59,34 +99,31 @@ presence.on("UpdateData", async () => {
...
-    episod = document.querySelectorAll("#showmedia_about_media h4");
-    epName = document.querySelector("h4#showmedia_about_name");
-    episode = episod[1].innerText + " - " + epName.innerText;
+    const videoTitle = document.querySelector(".ellipsis .text-link span"),
+    episod = document.querySelectorAll("#showmedia_about_media h4"),
+    epName = document.querySelector("h4#showmedia_about_name"),
+    episode = episod[1].innerHTML + " - " + epName.innerHTML;
 
-    var timestamps = getTimestamps(
+    const timestamps = presence.getTimestamps(
```
ririxidev: merge with const above dx

ririxidev (changes requested):
well on line 1 `var` -> `const`

ScarVite:
But the thing i don't get, is codacy only looking at changed lines?

because it never showed any of these issues before, especially with that "any" type stuff

ririxidev (commented):
Well, codacy is weird that's why we are moving to deepscan soon

ririxidev (approved):

Bas950 (approved):

PR #3681, opened 2021-05-08
WidgetStyle presence.

DiscordAnaxes:
Here some images of the presence working:


DiscordAnaxes:
> Site is not 2 months old.

Heya!

Regarding this our real domain is widget.style, but we haven't hosted the website on it yet. If we get the website back up and running on the widget.style domain (which is older than 2 months). Can I create another pr?

ririxidev:
Will see

PR #3710, opened 2021-05-12
WebtoonXYZ fix | Removed unused const

MareGraphics:


ririxidev:
yeah well, fix your fork

PR #3738, opened 2021-05-15
Add Busmiles Presence

CurtisSouthern:
Adds presence for Busmiles website ([busmiles.uk](http://busmiles.uk/))

### DIFF
--- a/websites/B/Busmiles/dist/metadata.json
+++ b/websites/B/Busmiles/dist/metadata.json
@@ -1,0 +1,29 @@
+{
+    "$schema": "https://schemas.premid.app/metadata/1.3",
+    "service": "Busmiles",
+    "author": {
+      "name": "Curtis#7652",
+      "id": "148182035704315904"
+    },
+    "category": "other",
+    "logo": "https://i.imgur.com/LrL2cEW.png",
+    "thumbnail": "https://i.imgur.com/wPA7hMc.png",
+    "version": "1.0.0",
+    "color": "#23649e",
+    "tags": [
+      "busmiles",
+      "bus-miles",
+      "bus",
+      "miles",
+      "uk",
+      "busmilesuk"
+    ],
+    "url": [
+      "busmiles.uk",
+      "buswanker.uk",
+      "shop.busmiles.uk"
+    ],
+    "description": {
+      "en": "Busmiles, The bus mileage system."
+    }
+  }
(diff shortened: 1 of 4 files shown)


ririxidev:
Heya @CurtisSouthern!
We are so sorry but we need to deny your pull request regarding this presence due of:
 - Low quality presence
   - Basically not a lot of DOM readings.
   - Not a lot to show on presence.
 - Issues with DeepScan.
 
Regards,
**ririxi**. PreMiD Reviewer Team.
Contact: mail@ririxi.dev, ririxi#2721.

CurtisSouthern:
Hello there @ririxidev,

> not a lot of DOM readings.
What do you mean by this?

> Not a lot to show on presence.
Instead of individually creating statements for all the pages, i've instead referenced a point in each page that says the same information. This means that it'd show the same amount of information regardless.

> Issues with DeepScan
I've taken a look at the DeepScan report, I can't seem to think of a possible fix for this. The variable is needed for the state and I don't see an alternative way of doing the "any" part.

Regards.

ririxidev:
> What do you mean by this?
That you don't access a lot DOM.

> Instead of individually creating statements for all the pages, i've instead referenced a point in each page that says the same information. This means that it'd show the same amount of information regardless.
So basically - this site is not qualified to be as presence then :/

> I've taken a look at the DeepScan report, I can't seem to think of a possible fix for this. The variable is needed for the state and I don't see an alternative way of doing the "any" part.
Obviously you can, no `var` and you can use different type for example `string` or `HTMLElement` etc 😉

CurtisSouthern:
> So basically - this site is not qualified to be as presence then :/

Why does this mean it doesn't qualify? All the information that would usually be displayed in other presences are displayed in this one too, i've just used a slightly different approach.

> Obviously you can, no var and you can use different type for example string or HTMLElement etc 😉

I don't think you understand what I mean, the variable "title" is needed for the state of the presence.

ririxidev:
> I don't think you understand what I mean, the variable "title" is needed for the state of the presence.

Yeah I do, you use type `any` and you can use other since you assign query selector function so basically it's `HTMLElement` type.

> Why does this mean it doesn't qualify? All the information that would usually be displayed in other presences are displayed in this one too, i've just used a slightly different approach.

I mean by this, that site doesn't have a lot to offer. You basically only show title of the subpage/page and nothing else :/

CurtisSouthern:
@ririxidev 

> I mean by this, that site doesn't have a lot to offer. You basically only show title of the subpage/page and nothing else :/

I don't see how this is means it does not qualify? There are several other presences that do exactly the same thing as mine. Could you give me some examples of things that make it qualify then because i can't seem to think why it doesnt?

PR #3745, opened 2021-05-17
Add support for the Crunchyroll Beta

safan41:
This could be pushed, but the one thing it lacks is the indication of seasons. Not sure what the best approach would be since the episode page in the beta has no clear mention of what season the episode is in.
image


ririxidev:
No drafts so I will close it for now :/

PR #3753, opened 2021-05-17
Fixed load name at presence

AhmedSabry957:
image
image


ririxidev (changes requested):
Bump version

ririxidev:
No response from author since last requested changes

PR #3760, opened 2021-05-17
Add buttons to player & browse pages for animepahe

MulverineX:
Also replaced querySelector usage so it should be more efficient.


ririxidev (changes requested):
I think you don't know but `getTimestamps` function is already in `Presence` class and you need to use this one :3

MulverineX:
Oh huh, must be recent-ish. I'll swap it out.

ririxidev:
~~just to point out~~ it's for 4/5 months or so

ririxidev (approved):

Bas950 (changes requested):
Upload proof of it working

MulverineX:
@Bas950 https://streamable.com/mg019s

Bas950 (approved):

PR #3778, opened 2021-05-20
Fix url problem

Enevevet:
The presence was not working on "fr.overleaf.com" and others because I specified the subdomain "www"


ririxidev:
Then do regex since overleaf.com -> www.overleaf.com

PR #3779, opened 2021-05-20
Fixed URL problems for Overleaf Presence

Enevevet:
The language is not that important in the Presence since the language is just the display language for Overleaf and doesn't change the rest. The Presence just needed to be triggered on some specific subdomains.

### DIFF
--- a/websites/O/Overleaf/dist/metadata.json
+++ b/websites/O/Overleaf/dist/metadata.json
@@ -4,22 +4,8 @@
     "name": "Enevevet",
     "id": "329669021043523594"
   },
-  "url": [
-    "www.overleaf.com",
-    "cs.overleaf.com",
-    "es.overleaf.com",
-    "pt.overleaf.com",
-    "fr.overleaf.com",
-    "de.overleaf.com",
-    "sv.overleaf.com",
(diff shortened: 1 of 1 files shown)


websites/O/Overleaf/dist/metadata.json
```diff
@@ -4,14 +4,29 @@
...
+    "de.overleaf.com",
+    "sv.overleaf.com",
+    "tr.overleaf.com",
+    "it.overleaf.com",
+    "cn.overleaf.com",
+    "no.overleaf.com",
+    "da.overleaf.com",
+    "ko.overleaf.com",
+    "ja.overleaf.com"
+  ],
```
ririxidev, line 22: Just make regexp 🤦

Bas950, line 22:
```suggestion
  "url": "www.overleaf.com",
  "regExp": "([a-z0-9-]+[.])*overleaf[.]com[/]",
```

Bas950 (changes requested):

ririxidev (approved):

Bas950 (approved):

PR #3780, opened 2021-05-20
Zoom: Add users in room & fix leaving bug

MrBartusek:
Fixes a bug when presence displays that user is still in a meeting when on the meeting end screen
image

It now shows count of the participants in a room
image

### DIFF
--- a/websites/Z/Zoom/presence.ts
+++ b/websites/Z/Zoom/presence.ts
@@ -86,7 +86,7 @@ function memberCount() {
   const counter = document.querySelector(
     ".footer-button__participants-icon > .footer-button__number-counter > span"
   );
-  if (counter == null) {
+  if (typeof counter === "undefined") {
     return null;
   } else {
     return Number(counter.innerHTML);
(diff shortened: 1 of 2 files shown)


websites/Z/Zoom/presence.ts
```diff
@@ -66,3 +81,14 @@ function videoEnabled() {
       .getAttribute("aria-label") != "start sending my video"
   );
 }
+
+function memberCount() {
+  const counter = document.querySelector(
+    ".footer-button__participants-icon > .footer-button__number-counter > span"
+  );
+  if (counter == null) {
```
ririxidev, line 89:
```suggestion
  if (counter === null) {
```

ririxidev (changes requested):

websites/Z/Zoom/presence.ts
MrBartusek: Don't think that's actually a good idea here, if for some reason a browser return `undefined` from `document.querySelector`  it will fail

ririxidev (approved):

Bas950 (approved):

PR #3792, opened 2021-05-22
Rythm: getPageletiable workaround & billing page support

ACertainCoder:
**Name**
Rythm

**Changes**
- Workaround for slowdowns caused by getPageletiable
- Support for the new getPageletiable which at least doesn't freeze Firefox anymore
- Fallback to custom getPageletiable for older extensions (and the current stable version)
- Added the billing management page (no screenshots because I don't have premium, but it should work because it doesn't need any information off that page)

**Screenshots**
grafik


ririxidev:
Will merge/review code when 2.2.4 will be released fully to everyone.

ACertainCoder:
> 
> 
> Will merge/review code when 2.2.4 will be released fully to everyone.

It could theoratically get merged already. The presence checks the extension version and currently always uses the fallback.

ririxidev (approved):
Ah lol

ririxidev:
Ah, bump version too lel

ACertainCoder:
> 
> 
> Ah, bump version too lel

Whoops, now it should be good to go.

ririxidev (dismissed):

ririxidev (approved):

ACertainCoder:
Sorry, one final fix. They seem to have changed the way on how the remaining time gets calculated.

ririxidev (changes requested):
image
fix bc I added new eslint rule

ACertainCoder:
Sorry but why do you guys force us to do things the non-standard way? I can agree that some general formatting rules are a good thing but it's simply too much.

First it only was
```js
let a = 1; // x
let b = 2;
let c = 3;

let a = 1, // ok
    b = 2,
    c = 3
```
which is somewhat ok, but now we have to explicetly remove brackets? Really!? It makes the code more and more unreadable / harder to understand.

grafik
grafik

ririxidev (approved):

websites/R/Rythm/presence.ts
```diff
@@ -483,3 +405,215 @@ function getPageletiable(js: string): Promise<string> {
...
+      // Add the view song button if buttons should be displayed
+      if (showButtons) {
+        presenceData.buttons.push({
+          label: (await strings).buttonViewSong,
+          url: apiData.referenceId
+        });
+      }
+
+      // Check if the song isn't paused
+      if (apiData.pause == false) {
```
ririxidev, line 524:
```suggestion
      if (!apiData.pause) {
```

websites/R/Rythm/presence.ts
```diff
@@ -483,3 +405,215 @@ function getPageletiable(js: string): Promise<string> {
...
+    presenceData.details = "Documentation";
+    presenceData.state = document.querySelector("h1").textContent;
+
+    presence.setActivity(presenceData);
+    presence.setTrayTitle(`Rythm ${separator} Documentation`);
+    return;
+  }
+
+  // Premium
+  if (path[0] == "premium") {
```
ririxidev, line 606: maybe you could use `switch(path[0])` instead of `if (path[0] == ...)` 🤔

websites/R/Rythm/presence.ts
```diff
@@ -483,3 +405,215 @@ function getPageletiable(js: string): Promise<string> {
...
+    presenceData.largeImageKey = "r1";
+    presenceData.details = "Premium";
+    if (hash == "#plans") presenceData.state = "Pricing and plans";
```
ririxidev, line 609:
```suggestion
    if (hash === "#plans") presenceData.state = "Pricing and plans";
```

websites/R/Rythm/presence.ts
```diff
@@ -483,3 +405,215 @@ function getPageletiable(js: string): Promise<string> {
...
+        if (djAmount > 0 || userAmount > 0) {
+          // Add user information
+          presenceData.state =
+            `${(await strings).dj}s: ${djAmount}` +
+            ` ${separator} ` +
+            `Users: ${userAmount}`;
+        }
+      } else {
+        // Check the idle time
+        if (idleStamp == 0) idleStamp = Date.now() / 1000;
```
ririxidev, line 581:
```suggestion
        if (idleStamp === 0) idleStamp = Date.now() / 1000;
```

websites/R/Rythm/presence.ts
```diff
@@ -483,3 +405,215 @@ function getPageletiable(js: string): Promise<string> {
...
+  const host = window.location.hostname,
+    path = window.location.pathname.split("/").slice(1),
+    hash = window.location.hash,
+    presenceData: PresenceData = {},
+    showUsernames = await presence.getSetting("usernames"),
+    showServerDetails = await presence.getSetting("server-details"),
+    showButtons = await presence.getSetting("buttons");
+
+  // Billing management
+  if (path[0] == "app" && path[1] == "billing") {
```
ririxidev, line 451: same like below?

websites/R/Rythm/presence.ts
```diff
@@ -407,19 +317,28 @@ async function checkAPIConnection(
...
-  repeatAll: string;
-  dj: string;
-  buttonViewSong: string;
+async function getAPIInfo(path: string[]): Promise<APIInfo> {
+  const version = presence.getExtensionVersion();
+  let namespace: apiNamespace = undefined,
+    socketUrl: string = undefined;
+
+  // Fallback mechanism for older extensions
+  if (version >= 224) {
```
ririxidev, line 330:
```suggestion
  if (version >== 224) {
```

ririxidev (changes requested):
🤔

ACertainCoder:
🤦‍♂️ Forgot these, one sec.

websites/R/Rythm/presence.ts
ACertainCoder: This would be a really big switch statement and might make it confusing, that's why I decided to go with the normal if statement

websites/R/Rythm/presence.ts
ACertainCoder: This operator doesn't exist

ririxidev: oops

ririxidev (dismissed):

Bas950 (changes requested):
Update fork

ririxidev:
oops, sowwy

.eslintrc.json
```diff
@@ -71,4 +71,4 @@
   "reportUnusedDisableDirectives": true,
   "noInlineConfig": true,
   "plugins": ["@typescript-eslint"]
-}
+}
```
ririxidev, line 74: oops

ACertainCoder:
This new version also hides nicknames by default for privacy reasons. Users can change that behaviour by enabeling "Show nicknames" in the settings.
grafik

.eslintrc.json
ACertainCoder: Do I need to change something here?

ririxidev: maybe reverting would be nice?

ACertainCoder: I had to fix a merge conflict. I'm not able to reverse that without recreating this conflict. At least I think so...

ririxidev: Okay then, it's only space tho so yeah

ririxidev (approved):
Thank you for making changes regarding our new eslint rules 😄

Bas950 (approved):

PR #3830, opened 2021-05-27
fix(Netflix): fix the video page

SlowLife1661:
Screenshots aren't from me so ¯\_(ツ)_/¯

image
image


websites/N/Netflix/presence.ts
```diff
@@ -367,3 +365,8 @@ presence.on("UpdateData", async () => {
     }
   }
 });
+
+function $(tag: string, attribute: string, is: string){
```
ririxidev, line 369: Maybe name it more human-friendly

ririxidev (changes requested):

SlowLife1661:
They changed something again, I will come up with other solution kek

PR #3844, opened 2021-05-29
Fix buttons type

ImRodry:
This makes it so that you can only have a maximum of 2 buttons and not infinite like previously. I only made the second element of the array optional because I assume you'd want to add at least one button when setting buttons, although passing an empty array would technically also be correct.


ririxidev:
thanks

ImRodry:
Damn that was fast xD

PR #3851, opened 2021-05-30
staff pages added (botsfordiscord.com) & errors fixed

br1ss:


ririxidev:
Presence is fixed in #3834. I will include your changes too.

PR #3856, opened 2021-05-30
Color-hex.com presence

MiepHD:
I have revised my presence and open a request again for color-hex.com.


websites/C/color-hex/dist/metadata.json
```diff
@@ -0,0 +1,22 @@
...
+    "name": "Laila",
+    "id": "636229819620655134"
+  },
+  "service": "color-hex",
+  "description": {
+    "en": "Color-hex gives information about colors.",
+    "de": "Color-hex gibt Informationen über Farben."
+  },
+  "url": "www.color-hex.com",
+  "version": "0.1",
```
br1ss, line 12: only 1.0.0 version

MiepHD: I changed it.

rxri (changes requested):
image
it's still a thing.
and version `1.0.0` not `1.0`

MiepHD:
Am I only allowed to do one DOM reading?

ririxidev:
> Am I only allowed to do one DOM reading?

You NEED to do more DOM readings or provide more functionality.

MiepHD:
I have several dom readings and cover nearly all of the funtionalities of the website. I had asked only because i was worried that i am not allowed to have more than one reading.

ririxidev:
You only do that 2 times

MiepHD:
Yes for two pages it made sense to get infos from within the pages but for most pages it is enough to read the url or the title

ririxidev:
So this presence will not be accepted until you will not get more DOM readings or so 😃

MiepHD:
Should I use DOM readings instead of using title and path to make the code more complicated?

MiepHD:
I can't find a rule in the Guidelines which says that I must have a certain number of DOM Readings. Could you please tell me where I can see this rule? Why do I have to have a certain amount of DOM Readings?

ririxidev:
This presence is just qualified for low quality at this point since presence is more likely "static". And I'm not going to continue this conversation anymore since I already told you what to do.

MiepHD:
Did you visit the site at any time?? There are no more functions on the site! Because there are no more functions I can't add functions to the presence.

MiepHD:
> This presence is just qualified for low quality at this point since presence is more likely "static". And I'm not going to continue this conversation anymore since I already told you what to do.

It's a mostly static site. So isn't it logically that the presence is also like that?

MiepHD:
@br1ss What do you think about this presence? Has it enough functionality?

ririxidev:
Sorry but `br1ss` is not Presence Reviewer. If you can't provide more functionality it means that site is not qualified. I'm sorry but we are accepting only "good" presences, not presences like yours rn.
If you want to report my actions, please do to Staff Coordinator (Bas950), his discord is Bas950#0950.

PR #3925, opened 2021-06-12
Added Presence for Indiachan

vahya:
Added presence for the imageboard Indiachan, I've tried to follow every guideline and the site isn't illegal or on a free domain.
Sorry for the previous pull, forgot to implement log and catalog view for the boards, now it's done.


Bas950 (changes requested):
There is not a lot to the presence, there is a lot of unneeded status messages in the presence, and upload proof of it working.

Not sure if this will be accepted or not

ririxidev:
Lol no

PR #3929, opened 2021-06-12
Drama.gg Presence

toki0179:
I believe all issues have been solved. If it is not high enough quality please describe why and also what I can do to improve the quality.

### DIFF
--- a/websites/D/Drama/dist/metadata.json
+++ b/websites/D/Drama/dist/metadata.json
@@ -17,6 +17,6 @@
     ],
     "url": "drama.gg",
     "description": {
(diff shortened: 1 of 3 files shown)


VocalZero:
Hint: Have a look at the issues DeepScan currently shows for your Presence.

toki0179:
> Hint: Have a look at the issues DeepScan currently shows for your Presence.

I will fix them, but I believe some of them aren't from my presence. Also is it really necessary that I use double quotes instead of single quotes?

ririxidev:
Yes, that's what it says.

PR #3966, opened 2021-06-17
AniKuru

ghost:


ririxidev:
Sorry but no, because it only supports video player

nodzba:
sure, it doesn't have much more to support other than that tho

ririxidev:
then yeah sorry :/

PR #3969, opened 2021-06-18
Update URLs for Sparked Host

hwalker928:
Sparked Host has changed some of their website URLs, thus I have updated the extension.


rxri (changes requested):
maybe just do regex instead of this massive url array ;-;

hwalker928:
It's not regex, but it is better than the massive array.

ririxidev:
I'm talking about metadata too

hwalker928:
@ririxidev See above changes.

rxri (approved):

ririxidev:
sorry for waiting

PR #3989, opened 2021-06-23
Added Radabot.net presence

bizojs:
Profile_RadaPresence


ririxidev:
We are so sorry, but we need to decline your presence due to:
 - Promoting dashboard/bot itself by presence when it's not allowed.
 - Not a lot of sites supported (for me it means that this site doesn't offer a lot).
 - Not a lot of DOM readings.

PR #3999, opened 2021-06-25
Update presence.ts

zaito-fr:
J'ai pris du temps donc dis moi si j'ai bien fait ou pas (I took time so tell me if I did well or not)


ririxidev:
Look, first:
 - Author of presence didn't allow you to change clientID
 - You forgot to bump version in metadata.
 - Eee, why are you changing structure of code of this presence? Just add things that you want to add
 - Presence must be in English not in your language.

PR #4009, opened 2021-06-26
Added RedeDark Presence

phillipicardoso:


ririxidev:
First;
- using presence for advertising
 - this site is not qualified for presence

PR #4011, opened 2021-06-27
Add Taiko Web Presence

### LINKED ISSUE #3702: Taiko Web | taiko.bui.pm
- [x] Is it a popular site?
- [x] Is the website older than 2 months? 
- [ ] Is the site locked to a specific country/region?
- [ ] Is the site a paid service? (e.g. Netflix, Hulu)
- [ ] Does the website feature NSFW content? (e.g. porn, etc...)
- [ ] Are you a donator/patron?
- [x] Do you acknowledge that coding presences is completely voluntary and may take time for your service to be added regardless of priority?

** 太鼓ウェブ - Taiko Web https://taiko.bui.pm/ ** EX https://youtu.be/-6RHAyghrPc

** Thinks that should be added are see that your in the menu and looking around for song and what song is playing in game when you start playing a song. If it 1p or 2p game that your in and if there something that interest you guys that wood be interesting to add than add it. **

**
Taiko Web **

vahya:
Resolves PreMiD/Presences#3702


vahya:
Screenshot 2021-06-27 183256
Screenshot 2021-06-27 183333
Screenshot 2021-06-27 183409
Screenshot 2021-06-27 183433
Screenshot 2021-06-27 183502
Screenshot 2021-06-27 183533

websites/T/Taiko Web/presence.ts
```diff
@@ -0,0 +1,114 @@
...
+        let songId = parseInt(url.substr(29));
+        let song = assets.songs.find(song => song.id === songId);
+        window.song = {};
+        window.song.title_lang = song.title_lang;
+        window.song.title = song.title;
+        window.song.category = song.category;
+      }
+      open.call(this, method, url);
+    }
+  }(XMLHttpRequest.prototype.open);`;
```
ririxidev, line 26: image
and btw you are not allowed to access website in this way. If you want to do it use `getPageletiable` and compare like this.

websites/T/Taiko Web/presence.ts
```diff
@@ -0,0 +1,114 @@
...
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+      largeImageKey: "taiko_logo",
+      startTimestamp: browsingStamp
+    },
+    presenceData1: PresenceData = {
+      largeImageKey: "taiko_logo",
+      smallImageKey: "taiko_logo",
+      startTimestamp: browsingStamp
+    },
```
ririxidev, line 39: I don't really understand what's the point of this.

websites/T/Taiko Web/presence.ts
```diff
@@ -0,0 +1,114 @@
...
+    {
+      label: "Play game",
+      url: `https://${document.location.hostname}`
+    }
+  ];
+
+  if (invite !== null) {
+    presenceData.details = "Waiting for other player to join ...";
+    presenceData.buttons.push({
+      label: "Join Invite",
```
ririxidev, line 100:
```suggestion
      label: "Join the game",
```

rxri (changes requested):

websites/T/Taiko Web/presence.ts
vahya: I don't think there's any other way to check which song is playing, the whole website is rendered on a canvas and constant use of `getPageletiable` makes the game lag a lot so it's better to avoid more uses of it.

websites/T/Taiko Web/presence.ts
vahya: These are two slide of the slideshow? Do you want me to convert them to array of two objects?

websites/T/Taiko Web/presence.ts
vahya: Also I am not making any outside requests, just capturing and reading the url of the requests that the site makes, so I guess that should make it ok?

ririxidev: If you will use `getPageletiable` outside `UpdateData` event it will not do any lag since it will only execute once.

websites/T/Taiko Web/presence.ts
ririxidev: Yeah.

websites/T/Taiko Web/presence.ts
ririxidev: > Also I am not making any outside requests, just capturing and reading the url of the requests that the site makes, so I guess that should make it ok?

and yeah, well no. you still cannot.

ririxidev: image

Bas950: yea just put it outside of updatedata and run it every x seconds. Updatedata runs every 100ms or so, thats why it's so laggy.

ririxidev: We need to add some option to change interval of it in v3 because presence doesn't need to be updated every 100ms in every situation.

vahya: [This guy has added a script too](https://github.com/PreMiD/Presences/blob/00c43cf1113b289782d757f77be3238471d83e26/websites/T/Town%20of%20Salem/presence.ts)

ririxidev: It's presence from last year where I wasn't reviewer yet. There are new rules and I strictly follow them.

ririxidev: And that's game related thing, because from what I see owner even added divs for PreMiD specifically to get this presence working so why shouldn't we allow it?

vahya: I see your pov, but here are my reasons why this part of the code should stay as it is:

- Showing which song is playing is an important feature for the presence
- There is no other way to tell which song is playing (I have tried and failed, there's no simple variable holding the current song in `window`, and the website is rendered on a canvas so no help from DOM as well)
- The added script isn't malicious and adds no significant overhead

If you don't want to create an exception that's okay.

ririxidev: Well I did found. Websocket gives you songID that you can later get from their API even difficulty.
image

vahya: Wait, does it work for single player as well? It was not in my browser

ririxidev: From what I see, yes.

vahya: `p2.socket.onmessage = e => console.log(e.data);`

Did you do this as well? The id returned isn't song id though

vahya: I am pretty sure this is the id values for other players online, there's a net play option for players to play together.

ririxidev: Lemme see later when I will have time, because there must be a way :)

Bas950: @ririxidev anything on this?

ririxidev: I'm busy

ririxidev: I will try to do it tomorrow before me going to vacations but I don't really promise. If I will not do that, presence will be allowed to be in store for now in its state.

theusaf:
Did some looking around on the page. It seems like when you start playing a game, a lot of useful information is put in the `debugObj.controller` variable (exposed to window), including the song data (`.selectedSong`)

websites/T/Taiko Web/presence.ts
theusaf: Did some looking around on the page. It seems like when you start playing a game, a lot of useful information is put in the `debugObj.controller` variable (exposed to window), including the song data (`.selectedSong`)

vahya: I guess this could work

vahya:
Working on a fix, will put a new PR when done

websites/T/Taiko Web/presence.ts
vahya: This did not work for multiplayer, so I gave up on this idea.

PR #4037, opened 2021-07-01
Add presence for zrytezene.thatcakeid.com.

TheClashFruit:
Examples:

hTtJKlLWma
rz5rdj57lI

Note:
> The website is still under development and registration is disabled


ririxidev:
I'm sorry but we don't accept websites when they are on subdomains. We have exceptions for example `my.id` etc, but in this situation, no.

PR #4042, opened 2021-07-02
ADN Fixed Once Again (More long term solution)

### LINKED ISSUE #3832: [Presence | Bug] Anime Digital Network
**Describe the Bug**
The status just show "Anime Digital Network" when watching a episode

**Steps to Reproduce**
1. Open https://animedigitalnetwork.fr/
2. Start an episode

**Expected Behavior**
The status should be updated with anime title, episode...

**Screenshots**
unknown (2)

**Environment**
- OS: Windows 10
- Browser: Chrome
- Version: 91.0.4472.77 (64-bit)

DarkVillager:
Resolves #3832

(When can i add myself as a contributor for this I've fixed it around 5 times now)

(Yes there are multiple screenshots of smth looking the same, but it isn't the behind the scenes code isn't the same.)
adn6
adn1
adn2
adn3
adn4
adn5

### DIFF
--- a/websites/A/AnimeDigitalNetwork/presence.ts
+++ b/websites/A/AnimeDigitalNetwork/presence.ts
@@ -7,7 +7,7 @@ const presence = new Presence({
   });
 
 presence.on("UpdateData", async () => {
-  const video: HTMLVideoElement = document.querySelector("video.vjs-tech"),
+  const video: HTMLVideoElement = document.querySelector("#adn-video-js_html5_api"),
     presenceData: PresenceData = {
       largeImageKey: "logo"
     };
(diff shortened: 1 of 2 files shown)


ririxidev:
add yourself as contributor and fix codacy.

websites/A/AnimeDigitalNetwork/presence.ts
```diff
@@ -7,21 +7,29 @@ const presence = new Presence({
...
+        title2 = title.content.replace(/[0-9]/g, " ").substr(6),
+        title3 = title2.length - 131;
+      if (title3 < 0) {
+        const title4 = title2.length - 112;
+        presenceData.details = title2.slice(0, title4);
+      } else presenceData.details = title2.slice(0, title3);
+
+      const timestamps = presence.getTimestamps(
+        Math.floor(video.currentTime),
+        Math.floor(video.duration)
```
Bas950, line 31: dont think you need to math floor it, and you could use getVideoTimestamps (or its called something like that at least)

Bas950 (changes requested):

websites/A/AnimeDigitalNetwork/presence.ts
ririxidev: image
ofc you need to

PR #4043, opened 2021-07-02
Fix regex on Sparked Host

hwalker928:
It turns out I did the regex wrong, resulting in the presence not working :/


rxri (changes requested):
idk where is version bump

hwalker928:
I did in fact + .1 to the versioning, but I forgot to pull beforehand, so I was just going from .0 to .1, not .1 to .2

rxri (approved):

PR #4049, opened 2021-07-03
Add anoBoy Presence

Loneth:


ririxidev:
I'm sorry but your presence is going to be denied due of;
 - This site is most likely nsfw because of ads
 - You don't support every site
 - Wrong semver
 - I know that we allow sites that are already 2 months old etc but I just don't see this site in our store.

PR #4054, opened 2021-07-04
Add Presence: The Backrooms Wiki

SammCheese:
Adds a Presence for [The Backrooms Wiki](http://backrooms-wiki.wikidot.com)
Proof of it working is Below

image
image
image

Please let me know if something isn't up to par, so I can fix it


ririxidev:
Hey @SammCheese,
We are sorry but after discussion with our other reviewers we decided to decline your presence.
 - Low quality presence (only show pageTitle & changes details to show on what page they are)
 - Low quality code
 
We know that we have presences like that in our store but we decided to decline these to keep up with quality so presences like will be just declined.

PR #4066, opened 2021-07-06
Update metadata.json from Deezer

SnoopieTheBest:
I think if the system are the same of here, the presence work better for people who the presence doesn't want work


ririxidev:
deezer.com redirects to www.deezer.com, unnecessary

PR #4068, opened 2021-07-06
Discord Türkiye Presence

elvesydev:
I made it for our own site


ririxidev:
Eee no. Low quality presence and site is not qualified.

PR #4081, opened 2021-07-09
Add Gogoanime's new domain gogoanime.pe to metadata.json

TheUndo:
## Changes made
- Added gogoanime.pe to the domain list
- Reordered domains to prioritize active domains
- Changed from green to the official yellow theme color

## Why?
Gogoanime added a new domain and I noticed the color was wrong

## Who am I?
I'm an admin on Gogoanime

### DIFF
--- a/websites/G/Gogoanime/dist/metadata.json
+++ b/websites/G/Gogoanime/dist/metadata.json
@@ -15,6 +15,9 @@
     "nl": "Bekijk anime online in het Engels. U kunt gratis series en films online bekijken met Engelse ondertiteling. Gratis download anime van hoge kwaliteit."
   },
   "url": [
+    "gogoanime.pe",
+    "gogoanime.vc",
+    "gogoanime.news",
     "gogoanime.io",
     "gogoanime.sh",
     "gogoanime.so",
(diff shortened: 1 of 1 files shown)


rxri (changes requested):
> Reordered domains to prioritize active domains

doesn't matter anyways so idk why you did that also there is a regexp

bump version

ririxidev:
ngl, it works fine
image
second, from what I know it's broken (presence itself) so there is no reason to add it to array list.

Undooooo:
> ngl, it detects fine
> image
> second, from what I know it's broken (presence itself) so there is no reason to add it to array list.

I know it detects and I know about the regex. I wanted to add another supported domain because people capt tagging me on discord with screenshots of the precense without the correct domains.

I just tested the presence and it works fine.

Why would you close this PR? It's simply a metadata change. I can revert the domain array order back to the original order.

ririxidev:
Look, presence to detect website uses regexp in metadata and ur api in presence file (https://github.com/PreMiD/Presences/blob/master/websites/G/Gogoanime/presence.ts#L100), there is no need to add another domain to array, it's already big.

Undooooo:
OK that's fine if you wish to drop this PR @ririxidev I'm not gonna push any further. I'll just live with people telling me that the presence list is whack. FYI I don't think there should be that many domains listed, only our active ones (gogoanim.pe, gogoanime.vc & gogoanime.news) should be listed.

ririxidev:
I'm actually wondering why ppl ask you when you are not even an author of presence lol. Just tell them that and don't respond them 🤷

Undooooo:
> I'm actually wondering why ppl ask you when you are not even an author of presence lol. Just tell them that and don't respond them 🤷

Yeah go wonder hey? I to wonder things like why you can't just add the domains. ¯\_(ツ)_/¯

ririxidev:
Happy? 87163ea079f60bd42dc258f4a822584856f33f60 ? :)

Undooooo:
> Happy? [87163ea](https://github.com/PreMiD/Presences/commit/87163ea079f60bd42dc258f4a822584856f33f60) ? :)

Yes very happy thanks.

PR #4083, opened 2021-07-09
Adding ADKami basic interactions

antharuu:
I just added a simple interaction for ADKami.

You can see what you are browsing as well as what you are looking at.

Being a French site with a French content, I put the texts only in French.

adkami-premid-3

adkami-premid-1

adkami-premid-2

### DIFF
--- a/websites/A/ADKami/dist/metadata.json
+++ b/websites/A/ADKami/dist/metadata.json
@@ -1,0 +1,26 @@
+{
+  "author": {
+    "name": "Antharuu",
+    "id": "121667021862600704"
+  },
+  "service": "ADKami",
+  "description": {
+    "fr": "Streaming d'Anime, Manga, Drama, Séries, Hentai",
+    "en": "Anime, Manga, Drama, Séries, Hentai Streaming"
+  },
+  "url": "adkami.com",
+  "version": "1.0.0",
+  "logo": "https://www.adkami.com/favicon2.png",
+  "thumbnail": "https://streaming-one.com/wp-content/uploads/2021/03/FireShot-Capture-1547-Accueil-ADKami-m.adkami.com_-1024x496.png",
+  "color": "#9bd300",
+  "tags": [
+    "anime",
+    "hentai",
+    "drama",
+    "series",
+    "streaming",
+    "adkami"
+  ],
+  "category": "anime",
+  "regExp": "www.adkami[.]com([\\S]+)"
+}
--- a/websites/A/ADKami/presence.ts
+++ b/websites/A/ADKami/presence.ts
@@ -3,9 +3,9 @@ const presence = new Presence({ clientId: "863104116502429717" });
 presence.on("UpdateData", async () => {
   const presenceData: PresenceData = {
     largeImageKey: "logo"
-  };
+  },
 
-  const path: string = document.location.pathname;
+   path: string = document.location.pathname;
   let cat: string = path.split("/")[1];
 
   if (cat === "drama-streaming") cat = "drama";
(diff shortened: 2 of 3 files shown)


websites/A/ADKami/presence.ts
```diff
@@ -0,0 +1,43 @@
...
+       descr = descrParts.join(" ");
+
+      presenceData.details = titleParts.join("-");
+      presenceData.state = `${descr} (${lang})`;
+
+      const bigImage: string = document.querySelector("#body > section > div.col-12.col-l-9.fiche-look > div.row.blocshadow > div > img.col-12.m-hidden").getAttribute("src");
+
+      presenceData.smallImageText = bigImage;
+
+      console.log(bigImage);
```
EncryptedDev, line 29: Why do you have this?

websites/A/ADKami/presence.ts
```diff
@@ -0,0 +1,43 @@
...
+  const presenceData: PresenceData = {
+    largeImageKey: "logo"
+  },
+
+   path: string = document.location.pathname;
+  let cat: string = path.split("/")[1];
+
+  if (cat === "drama-streaming") cat = "drama";
+
+  if (cat !== "") {
```
EncryptedDev, line 13: Checking if it is an empty string?

EncryptedDev (changes requested):
Hi, not much to this presence, quite basic indeed - support more endpoints?
Look at my review too

websites/A/ADKami/presence.ts
antharuu: Oh, indeed I forgot to remove this, it was for testing.

websites/A/ADKami/presence.ts
antharuu: Yes, the string is empty if you are on the home page.

websites/A/ADKami/dist/metadata.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "service": "ADKami",
+  "description": {
+    "fr": "Streaming d'Anime, Manga, Drama, Séries, Hentai",
+    "en": "Anime, Manga, Drama, Séries, Hentai Streaming"
+  },
+  "url": "adkami.com",
+  "version": "1.0.0",
+  "logo": "https://www.adkami.com/favicon2.png",
+  "thumbnail": "https://streaming-one.com/wp-content/uploads/2021/03/FireShot-Capture-1547-Accueil-ADKami-m.adkami.com_-1024x496.png",
```
EncryptedDev, line 14: Also, these both need to be on imgur

antharuu: Ah I didn't know. I'll update it soon then.

ririxidev:
I'm sorry but I need to decline this presence;
- presence doesn't have a lot to offer
- mistakes within JS (no consts, messing up with variables)
- no imgur links
- weird variable naming
- and more...

antharuu:
> * presence doesn't have a lot to offer

That's pretty subjective... what can you base that on? Would the code be too short and simple for you?

It just brings what a basic integration brings. What do you expect from a streaming site like this?

Be constructive at least on this point.

ririxidev:
No support for media player, no support for all subpages etc 😃

antharuu:
Well, the subpages are useless because the title and the number of the episode is included in the page name. What do you expect from this ?

As for the media player, it seems complicated to me for a site like this one which includes dozens (or more) media players from third party sites. Especially since it would not bring much...

ririxidev:
Well sorry then, no these features - no presence in store :/

antharuu:
It seems unjustified to me, but so be it. I'll do what you want if I have the time...

PR #4109, opened 2021-07-15
Add Animes Orion Presence

vahya:
Screenshot 2021-07-15 145427
Screenshot 2021-07-15 145502
Screenshot 2021-07-15 145256
Screenshot 2021-07-15 145330
Screenshot 2021-07-15 145405


Bas950 (changes requested):
add video timestamps etc?

vahya:
> add video timestamps etc?

Videos don't play on the website, they play on another domain which the user is redirected to.

Bas950:
> > add video timestamps etc?
> 
> Videos don't play on the website, they play on another domain which the user is redirected to.

then you should probs support that, not this. Cause this alone isn't enough.

PR #4111, opened 2021-07-15
Added Presence for InfoBlock.gq & tac-fm.ml

InfoBlock:
InfoBlock.gq is an Community Website
tac-fm.ml is an Online Radio


Bas950:
Both low quality presences. And they probs don't meet the requirements (for radio presences).

PR #4140, opened 2021-07-22
Fix Brick Hill Presence

okdargy:
uh so i did an oopsie and didn't test my presence and it broke

so now im fixing it

Without Fix:
image

With Fix:
image


SlowLife1661:
That's because you check if that's details if so you set activity to nothing xD just add `!`

Slowlife01 (approved):

Bas950 (approved):

PR #4157, opened 2021-07-28
fix(BBC, Viu)

Slowlife01:
# BBC

- [x] Bumped version
- [x] Tested

## **Typo fixes**
Well, nothing much just fixed some typos.

## Attachment
image

## Change log
**Updated** `1` string 
```diff
- Read Aricle
+ Read Article
```

# Viu

- [x] Bumped version
- [x] Tested

### Check if user is searching
Now check if user is truly searching for something.

## Attachment
image

### DIFF
--- a/websites/A/Amazon Music/presence.ts
+++ b/websites/A/Amazon Music/presence.ts
@@ -39,33 +39,38 @@ presence.on("UpdateData", async () => {
     newLang = await presence.getSetting("lang").catch(() => "en"),
     showPlaylist = await presence.getSetting("showPlaylist");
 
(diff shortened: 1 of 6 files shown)


websites/A/Amazon Music/presence.ts
```diff
@@ -187,4 +187,4 @@ presence.on("UpdateData", async () => {
     presenceData.largeImageKey = "logo";
     presence.setActivity(presenceData);
   }
-});
+});
```
Bas950, line 190: undo

Bas950 (changes requested):

PR #4211, opened 2021-08-05
Fixed Deezer

### LINKED ISSUE #4201: [Presence | Bug] Deezer
**Describe the Bug**

Deezer recently made a small change to the UI, which breaks the detection of the paused/playing state. It now shows as if you always have the track paused, even though it is playing.

From what I can understand by looking at the source code, the SVG icon for playing now has the class `class="svg-icon-group-btn is-highlight"`, and sadly no ID, which makes detection harder:(( There is an aria-label, but it is localized - I'm Swedish, so for me it is equal to `Pausa` ("Pause" in Swedish) when a track is playing and `Spela` ("Play" in Swedish) when a track is paused. A thought - comparing the track duration timestamp every second will do, but might be inefficient. (I'm not the developer of this RPC, so I'm not going to interfere!)

**Steps to Reproduce**

1. Open the Deezer web application
2. Play any track
3. The PreMiD RPC will be stuck on displaying the "Paused" icon.

**Expected Behavior**

PreMiD to show a "Playing" icon and how much time is left on the track in the RPC.

**Screenshots**

Here, I'm playing a track:
image

But it shows up like this in the RPC:¨

image

**Environment**

- OS: 
- Browser: 
- Version: 
  
OS: Windows 10

Browser: Opera GX

Version: LVL3 (core: 77.0.4054.275)
(based on Chrome 91.0.4472.164) (64-bit)

**Additional context**

None

DarkVillager:
Resolves #4201
- Changed to detect image path for pause image so it works for every language. (play & paused button names are different on every language)
image
image

### DIFF
--- a/websites/D/Deezer/presence.ts
+++ b/websites/D/Deezer/presence.ts
@@ -54,10 +54,14 @@ presence.on("UpdateData", async () => {
     albumLink = document.querySelector("div.marquee-content")
       .children[0] as HTMLAnchorElement;
 
-    const paused =
-      document.querySelector(
-        ".svg-icon-group-item:nth-child(3) .svg-icon-pause"
-      ) === null;
+    const paused2 =
+    document.querySelector("#page_player > div > div.player-controls > ul > li:nth-child(3) > button > svg > g > path").outerHTML;
+    let paused: boolean;
+    if (paused2 == '<path d="m5 2 18 10L5 22V2z"></path>') {
+      paused = true
+    } else {
+      paused = false
+    }
     currentTime = document.querySelector(
       "div.player-track > div.track-container > div.track-seekbar > div.slider.slider-autohide > div.slider-counter.slider-counter-current"
     ).textContent;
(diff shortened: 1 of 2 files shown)


websites/D/Deezer/presence.ts
```diff
@@ -54,10 +54,14 @@ presence.on("UpdateData", async () => {
...
       .children[0] as HTMLAnchorElement;
 
-    const paused =
-      document.querySelector(
-        ".svg-icon-group-item:nth-child(3) .svg-icon-pause"
-      ) === null;
+    const paused2 =
+    document.querySelector("#page_player > div > div.player-controls > ul > li:nth-child(3) > button > svg > g > path").outerHTML;
+    let paused: boolean;
+    if (paused2 == '<path d="m5 2 18 10L5 22V2z"></path>')
```
SlowLife1661, line 60:
```suggestion
    if (paused2 === '<path d="m5 2 18 10L5 22V2z"></path>') 
```

Slowlife01 (approved):

Bas950 (approved):

PR #4223, opened 2021-08-08
Deezer - Own mp3s now work for the presence.

### LINKED ISSUE #3906: [Presence | Bug] Deezer
**Describe the Bug**

The Deezer Discord Rich Presence does not update the Now Playing if you are playing one of your own, custom MP3s. It is stuck on the song that was played previously.
**Steps to Reproduce**

1. [Upload an MP3 to Deezer](https://support.deezer.com/hc/en-gb/articles/115004221605-Uploading-MP3s-to-Deezer)
2. Play a "regular song" from a playlist
3. Play the custom MP3
4. The "Now playing" will not update to the Custom MP3 title

**Expected Behavior**

Even though I don't expect the buttons on the RPC to work with a custom MP3 (since it is only visible to you and does not have a public artist link), I do expect the "Now playing" information to update to the title and the artist of the Custom MP3.

**Screenshots**

Playing a track that is not a custom MP3
image
Switching over to custom MP3 uploaded by me
image
RPC is still stuck on old song:
image

**Environment**

- OS: Windows 10
- Browser: Opera GX 
- Version: LVL2 (core: 75.0.3969.282) 
  

**Additional context**

marquee-content is still a relevant CSS selector, but artist links and album links are not available when using a custom MP3. I believe that is where the issue is. When using a custom MP3, the <a class="track-link"> elements are missing.

DarkVillager:
Resolves #3906
- Added check to see if the link for the song exists, if it doesnt then it shows a playbutton with a link for local uploaded songs.
image

### DIFF
--- a/websites/D/Deezer/presence.ts
+++ b/websites/D/Deezer/presence.ts
@@ -117,6 +117,7 @@ presence.on("UpdateData", async () => {
         .textContent.split(" · ")[0];
       showLink = albumLink = document.querySelector("div.marquee-content")
         .children[0] as HTMLAnchorElement;
+      if (!showLink) showLink = null;
       presenceData.details = title;
       presenceData.state = episode;
       presenceData.largeImageKey = "deezer";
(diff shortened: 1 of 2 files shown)


websites/D/Deezer/presence.ts
```diff
@@ -117,6 +117,7 @@ presence.on("UpdateData", async () => {
         .textContent.split(" · ")[0];
       showLink = albumLink = document.querySelector("div.marquee-content")
         .children[0] as HTMLAnchorElement;
+      if (!showLink) showLink = null;
```
Bas950, line 120:
```suggestion
```

websites/D/Deezer/presence.ts
```diff
@@ -127,13 +128,24 @@ presence.on("UpdateData", async () => {
...
       presenceData.endTimestamp = timestamps[1];
 
-      if (buttons) {
-        presenceData.buttons = [
-          {
-            label: (await strings).viewPodcast,
-            url: showLink.href
-          }
-        ];
+      if (showLink !== null) {
```
Bas950, line 131:
```suggestion
      if (showLink) {
```

Bas950 (changes requested):
that should work the same?

Bas950 (approved):

Slowlife01 (approved):

PR #4247, opened 2021-08-14
Created Disboard Presence

BrotRector:
# Disboard

This is a presence for Disboard. Disboard is a place where you can support your own or the Server from other people. All languages the page has are also displayed in the status.

image
image
image
image

### DIFF
--- a/websites/D/Disboard/dist/metadata.json
+++ b/websites/D/Disboard/dist/metadata.json
@@ -1,0 +1,22 @@
+  {
+    "$schema": "https://schemas.premid.app/metadata/1.3",
+    "service": "Disboard",
+    "author": {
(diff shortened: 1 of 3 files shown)


Slowlife01:
Heya, I sadly have to say that we've decided to **decline** your Pull Request, due to the following reasons:

- Low quality **Presence**.
  - Low quality code.
  - Doesn't read many DOM, basically show small amount of info.
- **Unqualified** site.
  - As this site is a "Discord server list" site and according to the [guidelines](https://docs.premid.app/en/dev/presence/guidelines), this kind of site have the extra requirements, and one of the the requirements says that the site must have a "very good" design and quality, but from my and others' perspective we think this site doesn't have a "very good" design.

PR #4257, opened 2021-08-16
Picarto Fixed

### LINKED ISSUE #3384: Picarto | picarto.tv
**Describe the Bug**
Container only says "Browsing..", does not show streams that are playing or play/paused status.

**Steps to Reproduce**
1. Go to picarto.tv
2. Open any stream

**Expected Behavior**
Expected to see the name of the streamer I'm watching in Discord RPC, however it's either only saying "Browsing.." or isn't picking up the page at all.

**Environment**
- OS: Windows 10 Pro 20H2
- Browser: Vivaldi
- Version: 3.6.2165.40 (Stable channel) (64-bit)
- PreMiD Extension Version: 2.2.0.7 beta
- PreMiD App Version: 2.2.0

**Additional context**
Picarto appears to have recieved a major update recently, so I'm not surprised it's not working as expected, but figured it would be good to let the devs of the container know.

DarkVillager:
- Picarto metadata updated,
- Picarto fully works again.
image
image

### DIFF
--- a/websites/P/Picarto/dist/metadata.json
+++ b/websites/P/Picarto/dist/metadata.json
@@ -1,20 +1,19 @@
 {
+  "$schema": "https://schemas.premid.app/metadata/1.3",
+  "service": "Picarto",
   "author": {
     "name": "TheSlowly",
     "id": "505071508851523599"
   },
-  "service": "Picarto",
+  "category": "videos",
+  "logo": "https://i.imgur.com/Oljre0e.png",
+  "thumbnail": "https://i.imgur.com/DSNnMk5.png",
+  "version": "1.2.8",
+  "color": "#ff0000",
+  "tags": ["video"],
+  "url": "picarto.tv",
   "description": {
     "en": "Picarto is the world’s leading live streaming platform for creative minds. Come join us and watch artists from all over the world during their creative process!",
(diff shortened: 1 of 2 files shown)


websites/P/Picarto/dist/metadata.json
```diff
@@ -1,20 +1,19 @@
...
-  "thumbnail": "https://i.imgur.com/DSNnMk5.png",
-  "color": "#FF0000",
-  "tags": [
-    "video"
-  ],
-  "category": "videos"
-}
\ No newline at end of file
+  }
+}
```
Bas950: there is no need to change the order...

Bas950 (changes requested):

websites/P/Picarto/dist/metadata.json
DarkVillager: There kinda is, it broke for me without the old metadata?

Bas950: bruh it's json... it doesn't matter.
You still just do metadata.author etc.

EncryptedDev:
🤨
> If a pull request is inactive for a period of 7 days, it will be promptly closed.

PR #4259, opened 2021-08-16
Indiachan - New Presence

natto1784:
A small yet epic imageboard.

Proofs:

### DIFF
--- a/websites/I/Indiachan/dist/metadata.json
+++ b/websites/I/Indiachan/dist/metadata.json
@@ -1,0 +1,32 @@
+{
+  "$schema": "https://schemas.premid.app/metadata/1.3",
+  "service": "Indiachan",
+  "author": {
+    "name": "natto",
+    "id": "874685117891739749"
+  },
+  "category": "other",
+  "logo": "https://i.imgur.com/hYySiGD.png",
+  "thumbnail": "https://i.imgur.com/EBTNFKT.png",
+  "version": "1.0.0",
+  "color": "#ef8e00",
+  "tags": [
+    "imageboard",
+    "anime",
+    "manga",
+    "politics",
+    "sports",
+    "discussions",
+    "forum",
+    "community"
+  ],
+  "url": [
+    "indiachan.com",
+    "www.indiachan.com",
+    "pajeet.top"
+  ],
+  "description": {
+        "hi_IN": "इंडियाचैन एक साधारण इमेजबोर्ड है जहां कोई भी किसी भी विषय के बारे में टिप्पणियां एवं चित्रों की सहायता से संवाद कर सकता है। आपको पोस्ट करने के लिए पंजीकरण करने की आवश्यकता नहीं है। इंडियाचैन विज्ञापनों या ट्रैकिंग का उपयोग नहीं करता है।",
+        "en": "Indiachan is a simple imageboard where anyone can post comments and share images about any topic. You do not need to register to post. Indiachan does not use ads or tracking. "
+  }
+}
--- a/websites/I/Indiachan/presence.ts
+++ b/websites/I/Indiachan/presence.ts
@@ -1,0 +1,61 @@
+const presence: Presence = new Presence({
+  clientId: "876813345699811328"
+});
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "logo"
+  }, p = document.location.pathname;
+  if (p === "/") {
+    presenceData.details = "Home";
+    presenceData.smallImageText = "Idling";
+    presenceData.smallImageKey = "dindu";
+    presence.setActivity(presenceData);
+    return;
+  }
+  if (p.startsWith("/a")) {
+    presenceData.details = "ありがとうございます";
+    presenceData.smallImageKey = "a";
+  } else if (p.startsWith("/g")) {
+    presenceData.details = "Install Gentoo";
+    presenceData.smallImageKey = "g";
+  } else if (p.startsWith("/b")) {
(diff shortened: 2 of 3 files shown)


websites/I/Indiachan/dist/metadata.json
```diff
@@ -0,0 +1,29 @@
...
+    "sports",
+    "discussions",
+    "forum",
+    "community"
+  ],
+  "url": "indiachan.com",
+  "description": {
+        "hi-IN": "इंडियाचैन एक साधारण इमेजबोर्ड है जहां कोई भी किसी भी विषय के बारे में टिप्पणियां एवं चित्रों की सहायता से संवाद कर सकता है। आपको पोस्ट करने के लिए पंजीकरण करने की आवश्यकता नहीं है। इंडियाचैन विज्ञापनों या ट्रैकिंग का उपयोग नहीं करता है।",
+        "en": "Indiachan is a simple imageboard where anyone can post comments and share images about any topic. You do not need to register to post. Indiachan does not use ads or tracking. ",
+        "pa-IN": "ਇੰਡੀਆਚੈਨ ਇਕ ਸਧਾਰਨ ਚਿੱਤਰਬੋਰਡ ਹੈ ਜਿੱਥੇ ਕੋਈ ਵੀ ਕਿਸੇ ਵੀ ਵਿਸ਼ੇ ਬਾਰੇ ਗੱਲ-ਬਾਤ ਅਤੇ ਚਿੱਤਰ ਪੋਸਟ ਕਰ ਸਕਦਾ ਹੈ। ਤੁਹਾਨੂੰ ਪੋਸਟ ਕਰਨ ਲਈ ਰਜਿਸਟਰ ਕਰਨ ਦੀ ਲੋੜ ਨਹੀਂ ਹੈ। ਇੰਡੀਆਚੈਨ ਵਿਗਿਆਪਨ ਜਾਂ ਟਰੈਕਿੰਗ ਦੀ ਵਰਤੋਂ ਨਹੀਂ ਕਰਦਾ।"
```
Slowlife01, line 27:
```suggestion
        "hi_IN": "इंडियाचैन एक साधारण इमेजबोर्ड है जहां कोई भी किसी भी विषय के बारे में टिप्पणियां एवं चित्रों की सहायता से संवाद कर सकता है। आपको पोस्ट करने के लिए पंजीकरण करने की आवश्यकता नहीं है। इंडियाचैन विज्ञापनों या ट्रैकिंग का उपयोग नहीं करता है।",
        "en": "Indiachan is a simple imageboard where anyone can post comments and share images about any topic. You do not need to register to post. Indiachan does not use ads or tracking. "
```

Slowlife01 (changes requested):

Slowlife01:
Also can you add more things?

natto1784:
I've already added all that can be added, there isn't really much more to an imageboard. If I can add more stuff in the future, I'll add it.

Slowlife01:
fix issues

natto1784:
fixed indentation

Slowlife01:
So is there more you can add?

natto1784:
there isn't more i can add, I've added all that i could. Also, cleaned up previous commits for clarity

websites/I/Indiachan/presence.ts
```diff
@@ -0,0 +1,61 @@
+const presence: Presence = new Presence({
+  clientId: "876813345699811328"
+});
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "logo"
+  }, p: string = document.location.pathname;
```
Slowlife01, line 8:
```suggestion
  }, p = document.location.pathname;
```

websites/I/Indiachan/presence.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (p === "/") {
+    presenceData.details = "Home";
+    presenceData.smallImageText = "Idling";
+    presenceData.smallImageKey = "dindu";
+    presence.setActivity(presenceData);
+    return;
```
Slowlife01, line 14: why set activity and return here

websites/I/Indiachan/presence.ts
```diff
@@ -0,0 +1,61 @@
...
+    presenceData.smallImageKey = "dindu";
+  } else if (p.startsWith("/yoga")) {
+    presenceData.details = "Fitness";
+    presenceData.smallImageKey = "yoga";
+  } else if (p.startsWith("/dhan")) {
+    presenceData.details = "Dhandho";
+    presenceData.smallImageKey = "dhan";
+  }
+
+  const parts: string[] = p.split("/");
```
Slowlife01, line 45:
```suggestion
  const parts = p.split("/");
```

websites/I/Indiachan/presence.ts
natto1784: to not make it go through the if else blocks below that

Slowlife01: well it's not like I don't know that but you check if the pathname is "/" so why does the code below matter other than the browsing one which is small image text anyways

natto1784: Because string functions like "includes" in the conditions are bound to make the code slower

Slowlife01: well, to my experience I wouldn't say so but sure if you say so, I won't argue

natto1784: Well I can just wrap the rest of the code in an else block if that's what you're asking me to do. But it won't be any different from using `return;` however.

Bas950 (approved):

Slowlife01 (approved):

PR #4272, opened 2021-08-19
Netflix UI update

### LINKED ISSUE #4228: [Presence | Bug] Netflix
**Describe the Bug**
When you watch a show or a movie, it does not show up on discord, it simply says ''browsing'' and the time elapsed.

**Steps to Reproduce**
This bug wasn't caused by anything I did to be reproduced, I simply went on Netflix and it said browsing. I have tried to re-install many times and also re-install the presence, and still doesn't work. I even turned ''browsing'' off but it still shows it.

**Expected Behavior**
My show was expected to show up on my discord with buttons, but instead it stays on ''browsing.''

**Screenshots**
Screenshot 2021-08-09 at 01 27 37

**Environment**

- OS: iOs (Mac)
- Browser: Chrome
- Version: 92.0.4515.131 (Official Build) (arm64)
  

**Additional context**
None

tedahn:
# Fixes: #4228 

Netflix Presence v4.2.0

Change querySelector to find HTML by `data-uid` instead of class. 
Fix applies to users with the new Netflix Web UI.

(Somewhat lazily) Tested on

Device:
- macOS Big Sur 11.5.2

Browsers: 
- Chrome
- Firefox
- Safari

Netflix Web UI Version: `10.15.7`

# Proof

## Before
Screen Shot 2021-08-19 at 6 04 03 AM

## After
Screen Shot 2021-08-19 at 6 09 35 AM
Screen Shot 2021-08-19 at 6 13 15 AM

### DIFF
--- a/websites/N/Netflix/presence.ts
+++ b/websites/N/Netflix/presence.ts
@@ -89,9 +89,9 @@ presence.on("UpdateData", async () => {
       ".watch-video--player-view video"
     );
     if (video && !isNaN(video.duration)) {
-      const showCheck = document.querySelector(
-          "[data-uia$='video-title'] span"
-        )
+      const showCheck = (document.querySelector(
+        "[class$='title'] .ellipsize-text span"
+      ) || document.querySelector("[data-uia$='video-title'] span"))
           ? true
           : false,
         timestamps = presence.getTimestampsfromMedia(video);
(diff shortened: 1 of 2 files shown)


Slowlife01:
Then what about the old UI... please use Logical OR

tedahn:
> Then what about the old UI... 
Not sure how to test that. 

Netflix uses 3 regions of AWS data centers but has an automated fail-safe that redirects traffic between them. Considering that, I wouldn't think they have different UI versions deployed to different regions because people would run into refreshing the page and seeing a different UI version when the fail-safe is triggered. But I could be wrong.

> please use Logical OR or Nullish coalescing operator. 
Any specific places you can point out?

> Afterward show proof of it working and bump version
Updating the PR

Slowlife01:
There are still people who haven't got the new UI so just use [Logical OR](https://mdn.io/Logical%20OR) or [Nullish coalescing operator](https://mdn.io/Nullish%20Coalescing%20operator) on these query selectors so it will work on both version and you don't have to test that

tedahn:
> There are still people who haven't got the new UI so just use [Logical OR](https://mdn.io/Logical%20OR) or [Nullish coalescing operator](https://mdn.io/Nullish%20Coalescing%20operator) on these query selectors so it will work on both version and you don't have to test that

Sure I can see that. It's done.

Slowlife01 (dismissed):

websites/N/Netflix/presence.ts
```diff
@@ -77,21 +77,21 @@ presence.on("UpdateData", async () => {
...
+   else if (oldLang !== newLang) {
     oldLang = newLang;
     strings = getStrings();
   }
 
   if (document.location.pathname.includes("/watch")) {
     const video: HTMLVideoElement = document.querySelector(
-      ".VideoContainer video"
+      ".watch-video--player-view video"
     );
```
Bas950, line 90: || or case. I have old UI and `.watch-video--player-view video` is not a thing yet for me.

Bas950 (changes requested):

websites/N/Netflix/presence.ts
tedahn: Oops. missed this one. Thanks for checking.

websites/N/Netflix/presence.ts
```diff
@@ -77,21 +77,19 @@ presence.on("UpdateData", async () => {
...
     oldLang = newLang;
     strings = getStrings();
   }
 
   if (document.location.pathname.includes("/watch")) {
-    const video: HTMLVideoElement = document.querySelector(
-      ".VideoContainer video"
-    );
+    const video: HTMLVideoElement = document.querySelector(".VideoContainer video")
+      ?? document.querySelector(".watch-video--player-view video");
```
Slowlife01, line 89: How about you just use
```js
document.querySelector("video")
```

Bas950: then it links the wrong element.

Slowlife01: Well I guess sure then

Slowlife01 (approved):

PR #4350, opened 2021-09-11
`getLevelInHeader` function fix, spelling corrections, added JP description translation.

upsetdog:
The bunpro.jp site had recent UI updates, and removed the `.navbar-user-level` class name. Which means the `getLevelInHeader()` function always throws an error, and prevents the Presence from continuing on with the `UpdateData` event.

As a quick fix, I have just added a null-check in the function (now it'll always return null, unless the `.navbar-user-level` class is ever re-added); however, I can't see the user's level reflected on every page, only their current XP and am not familiar with the XP->Level formula.

The current XP is just obtained with: `document.querySelector('.xp-current').innerText`

### DIFF
--- a/websites/B/Bunpro/dist/metadata.json
+++ b/websites/B/Bunpro/dist/metadata.json
@@ -8,7 +8,7 @@
   "category": "other",
   "logo": "https://i.imgur.com/NkfEDwV.png",
   "thumbnail": "https://i.imgur.com/uKMafFN.png",
-  "version": "1.0.2",
+  "version": "1.0.3",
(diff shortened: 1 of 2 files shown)


upsetdog:
Sorry, I forgot to update the `metadata.json` version from 1.0.2 to 1.0.3 😭😭

EncryptedDev (dismissed):

Slowlife01 (dismissed):

Slowlife01 (approved):

Bas950 (approved):

PR #4367, opened 2021-09-17
fix(GitHub Codespaces): small fixes

Snazzah:
This PR fixes some things:
- Workspace titles being incorrect (blaming #4276), goes back to using SCM tabs and using the `aria-label` attribute instead.
- Fixed selectors for preparing/inactive presences


Slowlife01 (approved):
Woops, was doing a quick workaround

Bas950 (approved):

PR #4385, opened 2021-09-20
add support for tweakers

bakzkndd:

PR #4388, opened 2021-09-21
Add support for TomsHardware

bakzkndd:
Please ignore the Tweakers part, it's been merged already (Pull request #4385)

**Added** Support for tomshardware.com, forum.tomshardware.com and tomshardware.com/uk

Proof of function:
image
image
image
image

### DIFF
--- a/websites/T/TomsHardware/presence.ts
+++ b/websites/T/TomsHardware/presence.ts
@@ -1,0 +1,98 @@
+const presence = new Presence({
+  clientId: "889780989386170420"
+});
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "tomshardware"
+  };
+
+  let pathname = document.location.pathname;
+
+  if (pathname.includes("uk/")) pathname = pathname.replace("uk/", "");
+  if (pathname.includes("uk")) pathname = pathname.replace("uk", "");
+
+  if (document.location.href.includes("news")) {
+    if (pathname.split("/")[2]) {
+      presenceData.details = "Reading an article...";
+      presenceData.state = `Article: '${pathname
+        .split("/")[2]
+        .replace(".html", "")}'`;
+    } else presenceData.details = "Scrolling through news articles...";
+  } else if (document.location.href.includes("reviews")) {
+    if (pathname.split("/")[2]) {
+      presenceData.details = "Reading a review...";
+      presenceData.state = `Review: '${pathname
+        .split("/")[2]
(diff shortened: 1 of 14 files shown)


Bas950:
> Please ignore the Tweakers part, it's been merged already (Pull request #4385)

Update your fork, so those files aren't shown in this PR.

websites/T/TomsHardware/presence.ts
```diff
@@ -0,0 +1,96 @@
...
+    presenceData.details = "Searching the site...";
+  } else if (pathname && pathname.length > 2) {
+    presenceData.details = "Browsing...";
+    presenceData.state = `On the page '${pathname.split("/")[1]}'`;
+  } else presenceData.details = "Visiting the front page...";
+
+  if (!presenceData.details) {
+    presence.setTrayTitle();
+    presence.setActivity();
+  } else presence.setActivity(presenceData);
```
Bas950, line 95:
```suggestion
  presence.setActivity(presenceData);
```
I think you can just do this because a state or detail is always set using the last to `if`/`else` cases

Bas950 (changes requested):

websites/T/TomsHardware/presence.ts
bakzkndd: I got that part from the documentation [here](https://docs.premid.app/en/dev/presence) under the **Getting started** part though

bakzkndd: nvm read that wrong, I can do that, you're right

websites/T/TomsHardware/presence.ts
```diff
@@ -0,0 +1,93 @@
+const presence = new Presence({
+  clientId: "889780989386170420"
+});
+
+presence.on("UpdateData", async () => {
+  const presenceData: PresenceData = {
+    largeImageKey: "tomshardware"
+  };
+
+  let { pathname } = document.location;
```
EncryptedDev, line 10: You are never reassigning the value of `pathname`, you should use `const` instead
```suggestion
  const { pathname } = document.location;
```

EncryptedDev (changes requested):

websites/T/TomsHardware/presence.ts
bakzkndd: CodeFactor failed when committing this.
Issue is that I am reassigning it, 2 lines below (L12 and L13), when it's a link with "/uk"

Bas950 (approved):

EncryptedDev (approved):

PR #4411, opened 2021-09-25
Added new minigame info.

Dzardys:


Bas950 (changes requested):
Update your fork, this presence already exists. (Different capitalization tho)

PR #4702, opened 2021-10-02
refactor:KeyLabler

EncryptedDev:
Removed the `keylabeler.yml` file for the following reasons:
- We don't use the `[Presence | Bug]` naming convention for issues as we use labels now
- We do not use the `[Presence | Feature Request]` naming convention as we use discussions now


ImRodry:
Not entirely within the scope of this PR but you should add a link in the ISSUE_TEMPLATE/config.yml file that sends users to discussions if they wanna make a feature request. Would make it more obvious imo

EncryptedDev:
> Not entirely within the scope of this PR but you should add a link in the ISSUE_TEMPLATE/config.yml file that sends users to discussions if they wanna make a feature request. Would make it more obvious imo

We have the pinned issue to do so & we can convert issues to discussions with a click of a button so it isn’t really needed

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,8 @@
 blank_issues_enabled: false
 contact_links:
-  - name: Official Discord server
+  - name: Discord Support
```
ImRodry, line 3: I think the old name was just fine tbh

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,8 @@
...
     url: https://discord.premid.app
-    about: Please use our ticket system for urgent requests!
+    about: If you are experiencing issues with your PreMiD, go to our Discord server for help.
```
ImRodry, line 5: Should still mention the ticket system, unless you want people to just message in a random chat which is also fine

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,8 @@
...
+  - name: Feature Requests & Presence Requests
+    url: https://github.com/PreMiD/Presences/discussions
```
ImRodry, line 7:
```suggestion
    url: https://github.com/PreMiD/Presences/discussions/new
```

ImRodry (changes requested):

.github/ISSUE_TEMPLATE/config.yml
EncryptedDev: Nope, as they need to copy the template before opening the discussion

ImRodry: Ah right, good luck with getting people to notice that
Might as well just link that discussion and have a link to the discussion creation page over there

PR #4749, opened 2021-10-13
fix(Plex): makes it work on library

Bas950:
image


StrikerFRFX:
Why are files from other presence fixes (which I think is included in the refractor) here?

Slowlife01 (approved):

PR #4828, opened 2021-10-27
Uploading Tio Anime extensión in Spanish

u0f:


EncryptedDev (commented):
These are just to list a few things which would improve the quality of this presence.

Think about utilising other features - e.g. presence settings, multiLang etc

PR #4832, opened 2021-10-27
fix: remove email from labels

StrikerFRFX:
All else works the same, only change is the removal of the email and `- Gmail` from this state:
Discord_nOL6olIydR


EncryptedDev (commented):
Upload proof that it is now working

Slowlife01 (approved):

EncryptedDev (approved):

PR #4882, opened 2021-11-09
Added Iris support

Dzardys:


Slowlife01:
Prettify it with our config and upload proof of the presence being working

Dzardys:
> Prettify it with our config and upload proof of the presence being working

Wdym saying "our config"? Little bit confused atm.

Slowlife01:
The config on the repo

Dzardys:
But I'm still not sure which file and how am I supposed to use, so I've used the same beautifier as always.

EncryptedDev:
> But I'm still not sure which file and how am I supposed to use, so I've used the same beautifier as always.

In the folder of the presence you made, run`prettier --write .` and it'll format to our repo standard

Slowlife01 (approved):

Bas950 (approved):

PR #4912, opened 2021-11-15
Re-add South Park URL's

NOTPIES:
adding again the deleted URL's because other countries use those URL's for south park so the URL is different for some countries and when those were deleted the SP presence stopped working on some countries unless you used an VPN

i attached an image of it working

image

image

### DIFF
--- a/websites/S/South Park/dist/metadata.json
+++ b/websites/S/South Park/dist/metadata.json
@@ -15,7 +15,7 @@
     "es": "South Park. Cada episodio de este programa galardonado y aclamado por las críticas está repleto de aventuras inolvidables, una sátira grosera y absurdidad graciosa."
   },
   "service": "South Park",
-  "version": "1.0.7",
+  "version": "1.0.6",
   "logo": "https://i.imgur.com/h6b3eEW.jpg",
   "thumbnail": "https://i.imgur.com/No75MzW.jpg",
   "color": "#b6122a",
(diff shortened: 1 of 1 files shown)


EncryptedDev (changes requested):
Bump version

websites/S/South Park/dist/metadata.json
```diff
@@ -4,14 +4,18 @@
...
+    "southpark.cc.com"
+  ],
   "description": {
     "en": "South Park. Every episode of this critically acclaimed and award-winning show is packed with unforgettable adventures, cutting satire, and hilarious absurdity",
     "nl": "South Park. Elke aflevering van deze veelgeprezen en bekroonde show zit boordevol onvergetelijke avonturen, scherpe satire en hilarische absurditeit",
     "es": "South Park. Cada episodio de este programa galardonado y aclamado por las críticas está repleto de aventuras inolvidables, una sátira grosera y absurdidad graciosa."
   },
   "service": "South Park",
-  "version": "1.0.7",
+  "version": "1.0.6",
```
EncryptedDev, line 18: _bump lmao_
```suggestion
  "version": "1.0.8",
```

NOTPIES: uhm im not sure thats an upgrade but more like an downgrade because 1.0.6 is the same as 1.0.8 then

Slowlife01 (approved):

EncryptedDev (approved):

PR #4893, opened 2021-11-11
feat: add eslint rules

EncryptedDev:
- Add `eslint-config-prettier` `eslint-plugin-prettier`
* Added `no-lonely-if` `prettier/prettier` `arrow-body-style` `prefer-arrow-callback` `@typescript-eslint/array-type`
- Updated presences to reflect this and bumped accordingly

*I feel like I've bumped too many presences as I messed masVer up, lmk if I've done something wrong*

PR #4918, opened 2021-11-16
Rewrite Duolingo from scratch

TheLukeGuy:
Hello! I noticed that the Duolingo presence wasn't nearly as good as it possibly could be, so I rewrote it (excuse the sloppy Photoshop job):

Before to After
([Full-res](https://i.imgur.com/40eeSrI.png) | [With website screenshots](https://i.imgur.com/BnwOZb6.png))

Specifically, this new version adds the following:

- Support for **many** more pages (literally everything I could find, from the basics like lessons/stories/shop/settings to more obscure pages like error pages and API responses)
- More information for pages that were already supported
- Support for all current (40) and future languages supported by Duolingo, compared to only 34 set languages supported by the old one
- Many general improvements to neatness
- A small image with the flag of the language you're learning and the name of the language when you hover over it
- Improved metadata with an updated thumbnail and color, URL regex for non-English subdomains, and descriptions for all languages that Duolingo has descriptions for
- A start timestamp which the old one didn't set for some reason

Unfortunately, this rewrite would require an exception to one of the contributing guidelines, specifically the one that disallows the use of `window.localStorage`. This property is read from once per second in the `updateLanguage` function:

```js
function updateLanguage() {
    const state = JSON.parse(window.localStorage.getItem('duo.state'));
    const courseId = state?.user?.currentCourseId;
    // ...
    const course = state.courses?.[courseId];
    // ...
}
updateLanguage();
setInterval(updateLanguage, 1000);
```

I've tried everything I could to get around this, but it looks like it's impossible to get the current Duolingo language on all pages without doing it this way. **I do not expect this exception to be granted**, but I figured it was worth a shot anyway. Without this, it wouldn't be possible (as far as I'm aware) to support all future languages, display the current language on most pages, or have that small image with the flag.

Besides that, everything should adhere to all of the contributor guidelines as far as I can tell. :smile:

### DIFF
--- a/websites/D/Duolingo/dist/metadata.json
+++ b/websites/D/Duolingo/dist/metadata.json
@@ -1,31 +1,51 @@
 {
   "$schema": "https://schemas.premid.app/metadata/1.5",
+  "service": "Duolingo",
   "author": {
-    "name": "TheSlowly",
-    "id": "505071508851523599"
+    "name": "Luke",
+    "id": "279622919242514432"
+  },
+  "category": "other",
+  "logo": "https://i.imgur.com/YKAf42L.png",
+  "thumbnail": "https://i.imgur.com/gVq8HiS.png",
+  "version": "2.0.0",
+  "color": "#58cc02",
+  "tags": ["education", "games", "languages"],
+  "regExp": "(www|(\\w{2}-)?\\w{2})\\.duolingo\\.(com|cn)",
+  "url": "www.duolingo.com",
+  "description": {
+    "ar": "‏الطريقة الممتعة والفعّالة لتعلّم لغة جديدة!",
+    "cs": "Bezplatný, zábavný a účinný způsob, jak se naučit jazyk!",
+    "de": "Effektiv und kostenlos eine Sprache lernen – und dabei Spaß haben!",
+    "el": "Ο δωρεάν, διασκεδαστικός και αποτελεσματικός τρόπος να μάθεις μια ξένη γλώσσα!",
+    "en": "The free, fun, and effective way to learn a language!",
+    "es": "¡La forma divertida, efectiva y gratis de aprender un idioma!",
+    "fr": "La méthode amusante, gratuite et efficace pour apprendre une langue !",
+    "hi": "भाषा सीखने का मज़ेदार, मुफ़्त और असरदार तरीका!",
+    "hu": "Egy ingyenes, szórakoztató és hatékony módja a nyelvtanulásnak!",
+    "id": "Cara gratis, seru, dan efektif untuk belajar bahasa!",
+    "it": "Il metodo gratuito, divertente ed efficace per imparare le lingue!",
+    "ja": "無料で楽しく外国語を学ぼう",
+    "ko": "재미있고 효과적인 무료 언어 학습 방법!",
+    "nl": "De gratis, leuke en effectieve manier om een taal te leren!",
+    "pl": "Darmowy, ciekawy i skuteczny sposób na naukę języka!",
+    "pt": "O jeito grátis, divertido e eficaz de aprender um idioma!",
+    "ro": "Modul gratuit, amuzant și eficient de a învăța o limbă!",
+    "ru": "Учите языки бесплатно, весело и эффективно!",
+    "th": "วิธีการเรียนภาษาที่ฟรี สนุกและมีประสิทธิภาพ!",
+    "tr": "Ücretsiz, eğlenceli ve etkili dil öğrenme yolu!",
+    "uk": "Безкоштовний, веселий та ефективний спосіб вивчення мови!",
+    "vi": "Cách học ngôn ngữ miễn phí, vui nhộn và hiệu quả!",
+    "zh": "学习外语，寓教于乐"
   },
   "contributors": [
+    {
+      "name": "TheSlowly",
+      "id": "505071508851523599"
+    },
     {
       "name": "narpi",
       "id": "321665259842830336"
     }
-  ],
-  "service": "Duolingo",
-  "description": {
-    "en": "Languages can be learned in a playful way. Duolingo is 100% free, fun and scientifically proven.",
-    "de": "Sprachen können auf spielerische Weise erlernt werden. Duolingo ist 100% kostenlos, macht Spaß und ist wissenschaftlich belegt.",
-    "nl": "Talen kunnen op een speelse manier geleerd worden. Dit spel is 100% gratis, leuk en wetenschappelijk bewezen.",
-    "pt_BR": "As línguas podem ser aprendidas de forma lúdica. Duolingo é 100% grátis, divertido e cientificamente comprovado."
-  },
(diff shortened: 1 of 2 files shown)


EncryptedDev:
Resolve ESlint issues

ConsoleLogLuke:
> Resolve ESlint issues

Done! Not sure how to make the `const`s at the top not look awful since they all need to be combined, but I tried. :/

websites/D/Duolingo/dist/metadata.json
```diff
@@ -1,31 +1,51 @@
...
+  "service": "Duolingo",
   "author": {
-    "name": "TheSlowly",
-    "id": "505071508851523599"
+    "name": "Luke",
+    "id": "279622919242514432"
+  },
+  "category": "other",
+  "logo": "https://i.imgur.com/YKAf42L.png",
+  "thumbnail": "https://i.imgur.com/gVq8HiS.png",
```
EncryptedDev, line 10: Use a screenshot of the homepage instead

websites/D/Duolingo/presence.ts
```diff
@@ -1,92 +1,232 @@
...
-      }
-      presenceData.details = "Taking a lesson";
-      presenceData.state = `Language: ${language}`;
-      presenceData.largeImageKey = "logo";
-      presence.setActivity(presenceData);
+  const languageId = course.learningLanguage;
+  if (FLAG_ICONS.includes(languageId)) language.imageKey = `flag_${languageId}`;
+  else if (languageId) language.imageKey = "flag_unknown";
+  else language.imageKey = "icon";
+  language.name = course.title ?? "nothing (yet!)";
```
EncryptedDev, line 80: Use a more apporopriate string instead of `nothing (yet!)`

websites/D/Duolingo/presence.ts
```diff
@@ -1,92 +1,232 @@
...
+}
+updateLanguage();
+setInterval(updateLanguage, 1000);
+
+function checkBasicPages(path: string[]) {
+  if (path.length === 0) {
+    const url = new URL(document.location.href);
```
EncryptedDev, line 87: bad practice to only use a variable once

websites/D/Duolingo/presence.ts
```diff
@@ -1,92 +1,232 @@
...
+    if (url.searchParams.get("isLoggingIn") === "true") {
+      presenceData.details = "Logging in";
+      return;
```
EncryptedDev, line 90: this for all these if statements
```suggestion
      return presenceData.details = "Logging in";
```

EncryptedDev (changes requested):

websites/D/Duolingo/dist/metadata.json
ConsoleLogLuke: That screenshot is the homepage when you're logged in, would you still prefer the landing page for when you're not logged in instead? (https://i.imgur.com/M3nbjZJ.png)

websites/D/Duolingo/presence.ts
ConsoleLogLuke: Okay I feel like that's highly debatable. That looks WAY more readable than:

```js
if (new URL(document.location.href).searchParams.get("isLoggingIn") === "true") {
```

And plus it keeps that line shorter. If you still insist on that change though, I'm fine with that.

websites/D/Duolingo/dist/metadata.json
EncryptedDev: It is up to you, I'd rather set the thumbnail as the home page screenshot as the dashboard doesn't look as good on the store

ConsoleLogLuke: Alright that's fine, I'll change it.

websites/D/Duolingo/presence.ts
EncryptedDev: We have made a ESLint plugin to prevent on-time-vars which will be implemented in #4893

ConsoleLogLuke: ughhh okay. I feel like that's a really bad decision and that it would be better to prioritize code readability over whatever else you're prioritizing, but that's not up to me so I'll change it.

websites/D/Duolingo/presence.ts
ConsoleLogLuke: Do you have any suggestions? I personally don't see a problem with `nothing (yet!)`, it indicates that the user hasn't chosen a language yet and it fits grammatically:

image

EncryptedDev: Set it to null

websites/D/Duolingo/presence.ts
```diff
@@ -1,92 +1,253 @@
...
+    presenceData.state = document.title;
+    return true;
+  }
+  if (API_ENDPOINTS.includes(path[0])) {
+    presenceData.details = "Viewing an API response";
+    presenceData.state = `Endpoint: /${path[0]}`;
+    return true;
+  }
+
+  return false;
```
Slowlife01, line 158: Use else if

ConsoleLogLuke: I can definitely do that if you want, but it makes no difference since I `return` in all of the `if`s anyway.

Slowlife01: style

websites/D/Duolingo/presence.ts
Bas950: > Okay I feel like that's highly debatable. That looks WAY more readable than:
> 
> ```js
> if (new URL(document.location.href).searchParams.get("isLoggingIn") === "true") {
> ```
> 
> And plus it keeps that line shorter. If you still insist on that change though, I'm fine with that.

in some cases like in ifs it might look uglier, but there are a LOT of "useless" variables in the presences. You can always add a comment above to if to clarify what it does, personally I think the if in your example is pretty readable, checking whether the search param has isLoggingIn to true. But since you are not the only contributor to this presence it might be nice to add a comment if you wish to do so.

ConsoleLogLuke: > in some cases like in ifs it might look uglier, but there are a LOT of "useless" variables in the presences.

No yeah I don't doubt that, but I feel like there are still a lot of cases where it's appropriate to create one-time variables to make the code more self-explanatory or to make it look nicer.

> You can always add a comment above to if to clarify what it does, personally I think the if in your example is pretty readable, checking whether the search param has isLoggingIn to true.

In that example I think the code is still self-explanatory, I just think it's harder to look at than before because everything is stuffed into a single line.

I just don't think it's a good idea to disallow one-time variables because there are many instances where it does clarify what code does and make it look much nicer. Yeah you could just use comments, but why do that when you can just create a separate variable for it which also has the added bonus of looking nicer?

EncryptedDev (commented):
Lgtm

EncryptedDev (approved):

Slowlife01 (approved):

PR #4919, opened 2021-11-16
Added new Iris sites support + some hotfixes

Dzardys:
In this update I just added some new sites support and fixed the possibility of using "/www/" in path, which could show you "/string.php" instead of just "string".


Dzardys:
As you can see, it works.
image
image
image

Slowlife01 (approved):

EncryptedDev (approved):

PR #4944, opened 2021-11-20
New Presence | Cinemana shabakaty

JUSTSAIF:
The website works only in Iraq

PR #4945, opened 2021-11-20
New Presence | Cinemana shabakaty

JUSTSAIF:
The website works only in Iraq


Slowlife01:
https://github.com/PreMiD/Presences/pull/4944#issuecomment-974686603

PR #4959, opened 2021-11-26
fix(Google Translate)

down-bad:
Fixed Google Translate and updated languages.

image

### DIFF
--- a/websites/G/Google Translate/presence.ts
+++ b/websites/G/Google Translate/presence.ts
@@ -4,6 +4,8 @@ const presence = new Presence({
 
 function languageCode(language: string): string {
   switch (language) {
+    case "auto":
+      return "Auto Detect";
     case "ar":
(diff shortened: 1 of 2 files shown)


Bas950 (approved):

EncryptedDev (approved):

websites/G/Google Translate/presence.ts
```diff
@@ -254,7 +266,7 @@ presence.on("UpdateData", async () => {
   const showType: boolean = await presence.getSetting("type");
 
   if (showType) {
-    presenceData.details = `Translating: ${typet}`;
+    presenceData.details = `Translating: ${type}`;
     presenceData.state = `From: ${languageCode(from)} - To: ${languageCode(
```
ImRodry, line 270: I would recommend removing the colons and the dash here to make it more readable

PR #4970, opened 2021-11-29
feat(Google Translate): settings, support more pages

down-bad:
Sorry for making a PR about this service again. Last commit had an occasional but common bug where it wouldn't show the source language.
So instead of just implementing a small fix, I took the chance to make use of query selectors instead of reading the data from the url (it was causing issues), add status for contributing and the about page, customizable status messages, and more languages.
Hopefully there won't be any more PRs about this anytime soon.

image

### DIFF
--- a/websites/G/Google Translate/dist/metadata.json
+++ b/websites/G/Google Translate/dist/metadata.json
@@ -26,7 +26,7 @@
   },
   "url": "translate.google.com",
   "regExp": "translate.google([.][a-z]+)+[/]",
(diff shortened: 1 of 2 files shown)


Bas950 (approved):

Bas950:
@ImRodry will be happy with this

ImRodry (approved):
Very happy yes (PR title should have more detail tho)

Slowlife01 (approved):

PR #4980, opened 2021-12-02
Fix Elements Path with New for JOOX Presence

ryukora:
Path Lines: 30, 37 44


nxzlem:
This is for fixing elements path on JOOX Presence only.

nxzlem:
The problem that causes errors and needs to fix:
!(https://files.catbox.moe/feol2s.png)[feol2s]

Slowlife01 (changes requested):
Bump version and show proof of it working

Slowlife01 (approved):

EncryptedDev (approved):

PR #5002, opened 2021-12-07
Add ArtStation presence

kyrie25:
#### I tried to cover as most pages and subdomains as possible for this but can still possibly miss some, but these are some  examples of how the presence works.  
  
Addressing discussion #4762 
* Home page  
image  
* Viewing an artwork  
image  
* Editing profile  
image  
* Checking inbox  
image  
* Learning  
image
image
* Viewing job postings
image
* Viewing a portfolio
image

### DIFF
--- a/websites/A/ArtStation/presence.ts
+++ b/websites/A/ArtStation/presence.ts
@@ -68,8 +68,10 @@ presence.on("UpdateData", () => {
     presenceData.details = "Editing profile";
   else if (document.location.href.indexOf("project/new") > -1)
     presenceData.details = "Uploading an artwork";
-  else if (document.location.hostname === "magazine.artstation.com")
+  else if (document.location.hostname === "magazine.artstation.com") {
     presenceData.details = "Reading magazines";
+    presenceData.state = document.title;
+  }
   else if (document.location.hostname == "www.artstation.com")
     presenceData.details = "Exploring artworks"
   else {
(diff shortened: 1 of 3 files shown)


websites/A/ArtStation/presence.ts
```diff
@@ -0,0 +1,80 @@
...
+    startTimestamp: browsingTimestamp
+  };
+
+  if (document.location.pathname.startsWith("/messages"))
+    presenceData.details = "Checking inbox";
+  // (not functional)
+    // else if (document.querySelector("user-profile")) {
+    // presenceData.details = "Viewing a profile";
+    // presenceData.state = document.querySelector('meta[property="og:title"]').getAttribute("content");
+  // }
```
Slowlife01, line 18: Why?

kyrie25: The only differentiation between a normal page and an artist's profile that I could find is this Class name part and my `querySelector` can't actually detect it and thus labels it as a normal page
image
I left it as a comment in the code hoping somebody else can find a fix for it, or I can delete it if it's not allowed/unnecessary

kyrie25:
Added several small image for `Checking inbox`, `Viewing an artwork`, `Exploring artworks`, `Learning` and `Learning` branches, `Editing profile`, `Uploading artwork`, `Viewing portfolio`
image
image

websites/A/ArtStation/presence.ts
kyrie25: Done

Slowlife01 (dismissed):

kyrie25:
Added a `View Artwork` button when viewing an artwork
image

EncryptedDev (approved):

Slowlife01 (approved):

Issue #4998, opened 2021-12-05
Stadia | stadia.google.com


Nalci:
### Description

I have a problem with the stadia presence. I installed it yesterday and when I'm on the stadia page it says on discord only "can't read page". When I'm in a game, it shows the game name correctly.

The problem exists also with PreWrap.

### Steps to Reproduce

1. Install the stadia presence
2. Go to stadia.google.com and Login
3. Go to home or browse the store and look at your status in Discord

### Expected Behavior

It should display the pages name where you on.

### Screenshots

![stadia_presence_1](https://user-images.githubusercontent.com/95542245/144763528-c0dcc77e-b110-4dcc-98b4-a0f7a3d463ba.JPG)

### Operating System (OS)

Windows 10

### Browser

Chrome

### Browser Version

Version 96.0.4664.45 (Offizieller Build) (64-Bit)

Pjoottr:
Sorry, I can't seem to reproduce this. Maybe try [refreshing your cache](https://support.google.com/accounts/answer/32050)?

Version 96.0.4664.93 (Official Build) (64-bit)

![image](https://user-images.githubusercontent.com/93800318/144932505-8c856326-cfc7-4b46-9688-91de41ee0c50.png)

Nalci:
Hello,

thank you for the reply!

I deleted all cache and data but nothing changed. I messed a bit aroung because it really wondered me why it worked for you but not for me.

I noticed some things:
1. If I have only one Google Account active, the URL is "https://stadia.google.com/home" and it works correctly
2. If I have multiple Google Accounts active and choose the one for Stadia, it changes to "https://stadia.google.com/u/1/home" - it shows "can't read page". But if I open "https://stadia.google.com/u/1/store" it shows "Viewing Stadia Store" correctly. Maybe it's a URL thing?
3. If I browse my library or click on discovery it shows again "can't read page". Is it maybe possible, to add also "Viewing Library" or "Browsing DIscovery" to it?

I hope it's not too confusing, english is not my native language.

Thank you very much!

Slowlife01:
What's the URL of the library?

Nalci:
The URL is https://stadia.google.com/library

PR #5003, opened 2021-12-07
Fixed issue #4998 and refactored, removed deprecated code

Pjoottr:
If you entered the site being a second user the URL structure would switch, from `stadia.com/home` to `stadia.com/u/1/home`, fixed by checking if the URL included home, then expect the user was at the home page, rather than looking for an exact URL match.

Before:
image

After:
image

### DIFF
--- a/websites/S/Stadia/dist/metadata.json
+++ b/websites/S/Stadia/dist/metadata.json
@@ -4,6 +4,12 @@
     "name": "Alexx",
     "id": "604453403640463371"
   },
+  "contributors": [
+    {
+      "name": "Pjoottr",
+      "id": "906309000465903697"
+    }
+  ],
   "url": "stadia.google.com",
   "description": {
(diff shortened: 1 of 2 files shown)


EncryptedDev:
Fixes #4998

websites/S/Stadia/presence.ts
```diff
@@ -10,10 +10,10 @@ let gameName: HTMLElement,
 presence.on("UpdateData", async () => {
   const presenceData: PresenceData = {
     largeImageKey: "logo",
-    startTimestamp: browsingStamp
+    startTimestamp: browsingStamp,
```
EncryptedDev, line 13: Revert this change

EncryptedDev (changes requested):
bump version

websites/S/Stadia/presence.ts
Pjoottr: All done 👍

EncryptedDev (commented):
You still need to bump the version in `metadata.json`

Pjoottr:
Oh yes, I'm sorry forgot about that. It would be 1.0.10 right?

websites/S/Stadia/dist/metadata.json
```diff
@@ -4,6 +4,12 @@
     "name": "Alexx",
     "id": "604453403640463371"
   },
+  "contributors": [
+    {
+      "name": "Pjoottr",
+      "id": "906309000465903697"
+    }
+  ],
```
Slowlife01, line 12: image
Not exactly a quarter yet if you add support for library path then sure

Slowlife01 (approved):

EncryptedDev (approved):

PR #5024, opened 2021-12-09
workflows: fix SV and delete dependabot-merge

EncryptedDev:
Workflow no longer works as the `main` branch is a protected branch. Reviewers can just approve Dependabot PRs from now on


EncryptedDev:
Closes #5020

Bas950 (approved):

Slowlife01 (approved):

.github/workflows/SV.yml
```diff
@@ -1,13 +1,9 @@
 name: Schema Validation Check
 on:
   push:
-    branches:
-      - main
     paths:
       - "**/metadata.json"
```
Slowlife01, line 5: I think you can just remove this to just make the check pass if there's no metadata file

EncryptedDev: That change makes sense, I’ll put that in place this evening. 

About it not working, it clearly does not. Look at any dependabot PR, always fails because the main branch is a protected branch

Slowlife01: Well for sure that's because it's doesn't contains any changes to metadata file since it only run if the PR contains the changes to metadata file

image

ImRodry: I have no clue what you're on about. The idea here is to make it so the workflow **doesn't** run on PRs that **don't** edit the metadata files, it won't produce any errors, it simply **doesn't run**

Bas950 (approved):

ImRodry:
mfw approves the approval

ImRodry (approved):

.github/workflows/SV.yml
Slowlife01: I mean make it so just that you won't have to force merge because the check is required

ImRodry: That’s the whole point of this

Slowlife01: Removing the `branches` field won't do anything, it will just make it so it can run on any branches

EncryptedDev: branches cheeks for /ref/head so the branch it comes from so it is not useful for our use

PR #5042, opened 2021-12-14
Add Stack Exchange Presence

avikrao:
Creating a presence for Stack Exchange and all of its subdomains and all the domains superuser.com, askubuntu.com, and serverfault.com, all of which are sites under Stack Exchange's network. Did NOT add support for Stack Overflow, as it already has a standalone presence.
image
image
image

### DIFF
--- a/websites/S/Stack Exchange/dist/metadata.json
+++ b/websites/S/Stack Exchange/dist/metadata.json
@@ -4,8 +4,8 @@
       "name": "AcidEpicice",
       "id": "249992970235936768"
     },
-    "url": ["stackexchange.com", "serverfault.com", "askubuntu.com", "superuser.com"],
-    "regExp": "(.*\\.)*(stackexchange\\.com|superuser\\.com|askubuntu\\.com|serverfault\\.com)",
+    "url": ["stackexchange.com", "serverfault.com", "superuser.com"],
+    "regExp": "(.*\\.)*(stackexchange\\.com|superuser\\.com|serverfault\\.com)",
     "description": {
       "en": "Stack Exchange is a network of question-and-answer websites on topics in diverse fields, each site covering a specific topic, where questions, answers, and users are subject to a reputation award process. The reputation system allows the sites to be self-moderating."
     },
--- a/websites/S/Stack Exchange/dist/metadata.json
+++ b/websites/S/Stack Exchange/dist/metadata.json
@@ -4,14 +4,14 @@
       "name": "AcidEpicice",
       "id": "249992970235936768"
     },
-    "url": "stackexchange.com",
-    "regExp": "(.*\\.)*stackexchange\\.com",
+    "url": ["stackexchange.com", "serverfault.com", "askubuntu.com", "superuser.com"],
+    "regExp": "(.*\\.)*stackexchange\\.com|superuser\\.com|askubuntu\\.com|serverfault\\.com",
     "description": {
       "en": "Stack Exchange is a network of question-and-answer websites on topics in diverse fields, each site covering a specific topic, where questions, answers, and users are subject to a reputation award process. The reputation system allows the sites to be self-moderating."
     },
     "service": "Stack Exchange",
     "version": "1.0.0",
-    "logo": "https://i.imgur.com/qHTJHmT.png",
+    "logo": "https://i.imgur.com/WCho2QO.png",
     "thumbnail": "https://i.imgur.com/Xax7CJQ.png",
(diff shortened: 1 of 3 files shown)


websites/S/Stack Exchange/presence.ts
```diff
@@ -0,0 +1,76 @@
+const presence = new Presence({
+    clientId: "919817726195814431"
+  }),
+  startTimestamp = Math.floor(Date.now() / 1000);
```
EncryptedDev, line 4:
```suggestion
  browsingTimestamp = Math.floor(Date.now() / 1000);
```

websites/S/Stack Exchange/presence.ts
```diff
@@ -0,0 +1,76 @@
...
+    presenceData.largeImageKey = "askubuntu";
+    presenceData.details = "Ask Ubuntu Meta";
+  } else if (hostname === "superuser.com") {
+    presenceData.largeImageKey = "superuser";
+    presenceData.details = "Super User";
+  } else if (hostname === "meta.superuser.com") {
+    presenceData.largeImageKey = "superuser";
+    presenceData.details = "Super User Meta";
+  } else {
+    const subStack = document.querySelector("meta[property='og:site_name']"),
```
EncryptedDev, line 33: Variable is only used once

websites/S/Stack Exchange/presence.ts
```diff
@@ -0,0 +1,76 @@
...
+        "superuser.com",
+        "meta.superuser.com",
+        "askubuntu.com",
+        "meta.askubuntu.com"
+      ].includes(hostname)
+    )
+      presenceData.state = "Main Page";
+    else presenceData.details = "Main Page";
+  } else if (pathname.includes("/questions")) {
+    const titleElem = document.querySelector(".question-hyperlink");
```
EncryptedDev, line 59: Variable is only used once

EncryptedDev (changes requested):

websites/S/Stack Exchange/dist/metadata.json
```diff
@@ -0,0 +1,26 @@
+{
+    "$schema": "https://schemas.premid.app/metadata/1.5",
+    "author": {
+      "name": "AcidEpicice",
+      "id": "249992970235936768"
+    },
+    "url": ["stackexchange.com", "serverfault.com", "askubuntu.com", "superuser.com"],
+    "regExp": "(.*\\.)*stackexchange\\.com|superuser\\.com|askubuntu\\.com|serverfault\\.com",
```
ImRodry, line 8: this seems rather redundant, unless you want to match all of those domains with anything before them, in which case the regex is incorrect
```suggestion
    "regExp": "(.*\\.)*stackexchange\\.com",
```

websites/S/Stack Exchange/dist/metadata.json
```diff
@@ -0,0 +1,26 @@
...
+    "color": "#1E5397",
+    "tags": [
+      "help",
+      "community",
+      "discussion",
+      "forum",
+      "question",
+      "answer"
+    ],
+    "category": "other"
```
ImRodry, line 25: move above tags pls

ImRodry (changes requested):

websites/S/Stack Exchange/dist/metadata.json
AvikRao: Sorry, could you elaborate what the issue with the regex is? If you mean that I'm incorrectly catching subdomains, I _do_ want to catch all subdomains of the listed sites

ImRodry: This regex will match `ggerhgyuierhbgre.stackexchange.com` but not `ergergerger.serverfault.com`, which is what you want

AvikRao: Looks like it does, actually?
image

ImRodry: It shouldn’t then, here’s the correct one (do test this though)
```suggestion
    "regExp": "$(.*\\.)*(stackexchange\\.com|superuser\\.com|askubuntu\\.com|serverfault\\.com)^",
```

Slowlife01 (commented):
There's already a presence for AskUbuntu

AvikRao:
> There's already a presence for AskUbuntu

removed askubuntu

Slowlife01 (approved):

EncryptedDev (approved):

PR #5049, opened 2021-12-16
Improve input field selector for Google

Sominemo:
Current input field selector is complex and unreliable due to autogenerated class names that Google uses (for example, the presence doesn't work for me due to homepageInput being null, spamming my console with the error). Instead, let's use input field's constant properties.


Slowlife01 (changes requested):
Bump version and show proof of the presence working

Sominemo:
Proofs
image
image

Slowlife01 (approved):

EncryptedDev (approved):

PR #5054, opened 2021-12-17
Add CS:GO Stash presence

kyrie25:
Examples of how the presence works:
image
image
image
image
image
image

### DIFF
--- a/websites/C/CSGO Stash/presence.ts
+++ b/websites/C/CSGO Stash/presence.ts
@@ -1,0 +1,182 @@
+const presence = new Presence({
+    clientId: "918832169823125555"
+  }),
+  browsingTimestamp = Math.floor(Date.now() / 1000);
+
+presence.on("UpdateData", () => {
+  const presenceData: PresenceData = {
+      largeImageKey: "logo",
+      startTimestamp: browsingTimestamp
+    },
+    shortTitle = document.title.substring(
+      0,
+      document.title.lastIndexOf("-") - 1
+    ),
+    path = document.location.pathname;
+
+  if (path.startsWith("/skin")) {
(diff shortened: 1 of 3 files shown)


EncryptedDev (changes requested):
Use `.textContent` instead of `.innerHTML` on the `document.querySelector()`s

websites/C/CSGO Stash/presence.ts
```diff
@@ -0,0 +1,182 @@
...
+    }`;
+    presenceData.smallImageKey = "items";
+    presenceData.smallImageText = "Viewing items";
+    presenceData.buttons = [{ label: "View Item", url: document.URL }];
+  } else if (path.indexOf("google-search") > -1) {
+    presenceData.details = "Searching for:";
+    presenceData.state = document.querySelector(
+      "#___gcse_0 > div > div > div > div.gsc-wrapper > div.gsc-resultsbox-visible > div > div > div.gcsc-more-maybe-branding-root > a > div > span.gcsc-find-more-on-google-query"
+    ).textContent;
+  } else if (document.location.href === "https://csgostash.com")
```
ImRodry, line 178: This is always true
```suggestion
  } else
```

kyrie25: @ImRodry not quite
It would not return as the case above when browsing  different sites
image

kyrie25: image
This is the home page case as above

kyrie25: If I recall correctly
`.href` also returns slashes and branches in the site URL so the usage of `.href` there is justified

Slowlife01 (approved):

EncryptedDev (approved):

PR #5086, opened 2021-12-22
feat(SoundCloud): Cover as largeImageKey, non async-y strings

Bas950:
image

### DIFF
--- a/websites/S/SoundCloud/presence.ts
+++ b/websites/S/SoundCloud/presence.ts
@@ -30,13 +30,7 @@ const presence = new Presence({
 
 let elapsed = Math.floor(Date.now() / 1000),
   prevUrl = document.location.href,
-  strings: {
-    play: string;
-    pause: string;
-    browse: string;
-    search: string;
-    listen: string;
-  } = null,
+  strings: Awaited<ReturnType<typeof getStrings>> = null,
   oldLang: string = null;
 
 const statics = {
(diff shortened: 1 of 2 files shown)


Slowlife01 (dismissed):
kek

websites/S/SoundCloud/presence.ts
```diff
@@ -154,97 +177,106 @@ presence.on("UpdateData", async () => {
...
+
+    presenceData.startTimestamp = startTimestamp;
+    presenceData.endTimestamp = endTimestamp;
+
+    if (cover) {
+      presenceData.largeImageKey = document
+        .querySelector<HTMLSpanElement>(
+          ".playbackSoundBadge__avatar.sc-media-image > div > span"
+        )
+        .style.backgroundImage.match(/"(.*)"/)[1]
```
Slowlife01, line 189: Actually, it could be null

Bas950: Actually, it cannot be null, this is inside of the if case which checks whether something is playing or not, which means the bar at the bottom of the screen is shown.

Slowlife01: *Some don't have cover*

Bas950: got an example?

Slowlife01: https://soundcloud.com/user48301249/vvkwvzuw2psc?in=user48301249/sets/heart

EncryptedDev (approved):

Slowlife01 (approved):

PR #5089, opened 2021-12-22
Updated animesoul and added shoob.gg. Sites split

Gib-git:


Gibigbig:
Will edit and make a new PR

PR #5094, opened 2021-12-23
feat(util): add and improve scripts

ImRodry:
This PR adds 2 new scripts and improves existing ones to use them.

- Refactored the package.json file to remove the :compile scripts and move them to their original script as they weren't used anywhere else anymore
- Updated SV to check if all Presences were bumped correctly. This currently has a flaw due to how the script was originally made which is that it will not produce an error if the PR didn't change the metadata file at all. This could be improved in the future.
- Added a script to compile all presences 1 by 1. This wasn't added to any CI so it can only be ran by maintainers when needed as it takes a long time to run. The script **must** compile each presence after the other, otherwise your computer will catch on fire
- Added script to sort metadata files according to the order on the docs and updated SE to run this as well.
- Did a small refactor on MV to remove the main function as it was redundant, everything else stays the same

### DIFF
--- a/util/tools/massCompile.ts
+++ b/util/tools/massCompile.ts
@@ -1,0 +1,84 @@
+import "source-map-support/register";
+
+import { exec as processExec } from "node:child_process";
+import { readFileSync as readFile, writeFileSync as writeFile } from "node:fs";
+import { promisify } from "node:util";
+import { green, red } from "chalk";
+import { sync as glob } from "glob";
+
+const exec = promisify(processExec);
+
+/*  NOTE: THIS IS A TOOL THAT IS ONLY MEANT TO BE USED
+    BY THE DEVS AND REVIEWERS FOR DEPLOYMENT PURPOSES,
+    PLEASE DON'T COMPILE OR RUN IT BEFORE MAKING A PULL
+    REQUEST UNLESS YOU'VE BEEN EXPLICITLY INSTRUCTED BY
+    A DEV TO DO SO, WHICH WILL MOST LIKELY NEVER HAPPEN.  */
+
+export const read = (path: string): string =>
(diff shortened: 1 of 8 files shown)


util/tools/metadataSorter.ts
```diff
@@ -0,0 +1,140 @@
...
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export const read = (path: string): string =>
+    readFile(path, { encoding: "utf8" }),
+  write = (path: string, code: Metadata): void =>
+    writeFile(path, JSON.stringify(code, null, 2), {
```
Bas950, line 22: why is that spacing so bork

util/tools/massCompile.ts
```diff
@@ -0,0 +1,84 @@
...
+/*  NOTE: THIS IS A TOOL THAT IS ONLY MEANT TO BE USED
+    BY THE DEVS AND REVIEWERS FOR DEPLOYMENT PURPOSES,
+    PLEASE DON'T COMPILE OR RUN IT BEFORE MAKING A PULL
+    REQUEST UNLESS YOU'VE BEEN EXPLICITLY INSTRUCTED BY
+    A DEV TO DO SO, WHICH WILL MOST LIKELY NEVER HAPPEN.  */
+
+export const read = (path: string): string =>
+    readFile(path, { encoding: "utf8" }),
+  write = (path: string, code: Metadata): void =>
+    writeFile(path, JSON.stringify(code, null, 2), {
```
Bas950, line 20: same here

util/tools/metadataSorter.ts
ImRodry: It's... not?

Bas950 (approved):

EncryptedDev (approved):

PR #5099, opened 2021-12-23
feat(YouTube): Add thumbnails as image keys

Snazzah:


Bas950 (changes requested):
I don't think this is a good one, at least make it a setting. But a no from me :/

Bas950 (changes requested):
bump schema version to 1.6

Bas950 (approved):

EncryptedDev (approved):

PR #5122, opened 2021-12-25
fix(scripts): metadata version and duped presence validation

ImRodry:
This PR fixes issues in two scripts:

- SV: fixes an issue where a presence would be checked twice if it was created and modified. Now all presences will only be checked once
- MS: correctly fetches the latest metadata version instead of hardcoding version 1.5 (why did I do that)

It also removes the PR template from the workflows file cuz why was it there to begin with

<details>
<summary> Proof to proving the creation/modification working </summary>
<br>

no

</details>


ImRodry:
Note that MS currently has a flaw which is that it will run schemaUpdater when calling latestMetadataSchema. This isn't really an issue though since we want that to run anyways, but it might impact performance or be an issue in the future. Maybe the code in schemaUpdater could be removed so that we only keep the function that fetches the version, and use that on MS only. If you want me to go with that option, let me know and I'll update the PR

ImRodry:
Deleted schemaUpdater and moved the function over to metadataSorter. If you want we could find a better name for that script now, or just merge like this

Bas950 (approved):

EncryptedDev (approved):

PR #5136, opened 2021-12-26
feat(scrap.tf): add presence 

Ali-Joon:
sc2
sc4
I think I should make some changes after this...


EncryptedDev (changes requested):
Rebase your branch

PR #5171, opened 2021-12-28
fix(Wikipedia)

### LINKED ISSUE #5115: Wikipedia | wikipedia.org
### Description

I use wikipedia a lot, quite more than most normal people using premid, and I assume that's why no one has come across the issue that I am having. When editing wikipedia in source mode, there are little buttons next to each section that say "edit source." (See image)
image
I have a setting enabled in wikipedia to show this button on the top of the article, next to it's title. (See image)
image
This causes an issue in premid where it includes the "[edit source]" in the title of the page, which is inaccurate. (See image)
image

### Steps to Reproduce

1. Sign into/create a wikipedia account if you do not already have one (required to set preferences in wikipedia)
2. Click the preferences button on the top right of the screen
3. Click on the section entitled "Gadgets"
4. Scroll to the "Appearance" section
5. Enable the first setting, which says "Add an [edit] link for the lead section of a page"
6. Press save at the bottom of the screen
7. Navigate to any wikipedia page
8. View premid status

### Screenshots

_No response_

### Expected Behavior

The title of the wikipedia page would show, according to the site

### Actual Behavior

The title of the page shows, along with extra text

### Page URL

https://en.wikipedia.org/wiki/Example

### Operating System (OS)

Windows 10

### Browser

Firefox

### Browser Version

95.0.2

kyrie25:
#### Resolves #5115 

<details>
<summary> Proof of the modification working </summary>
<br>

- Before:
image
- After:
image
- Does **not** break the presence when the user is not logged in/has not toggle the settings:
image
</details>

### DIFF
--- a/websites/W/Wikipedia/dist/metadata.json
+++ b/websites/W/Wikipedia/dist/metadata.json
@@ -163,7 +163,7 @@
   },
   "url": "wikipedia.org",
   "regExp": "([a-z0-9-]+[.])+wikipedia[.]org[/]",
-  "version": "2.7.12",
+  "version": "2.7.13",
   "logo": "https://i.imgur.com/4OX6OZi.png",
   "thumbnail": "https://i.imgur.com/vJiuzZy.png",
   "color": "#f9f9f9",
--- a/websites/W/Wikipedia/presence.ts
+++ b/websites/W/Wikipedia/presence.ts
@@ -113,6 +113,7 @@ const updateCallback = {
         details[
           [...document.querySelector("body").classList]
             .filter(v => /ns--?\d/.test(v))[0]
+            .split("[", 1)[0]
             .slice(3)
         ] || "Viewing a page"
       );


Slowlife01 (approved):

nyakase (changes requested):
I wouldn't really call this a "fix", as if you have other settings that insert stuff in the article's title, that would _also_ pop up.

There's a function you can use, `mw.config.get('wgPageName')`<sup>[1](https://www.mediawiki.org/wiki/Manual:Interface/JavaScript#mw.config)</sup>, which will get the page name for you without other things.

kyrie25:
@Vukkyy 
> I wouldn't really call this a "fix", as if you have other scripts that insert stuff in the article's title, that would _also_ pop up.  

`fix` is just a mandatory commitlint commit message template AFAIK, it's not my choice
> There's a function you can use, mw.config.get('wgPageName')1, which will get the page name for you without other things.  

The whole structure of this script is unfamiliar to me, and it uses functions to extract data in ways I have not spent that much time looking into yet. I only patched the problem that has been reported. I might be able do some more research and change it, but this is the solution that one of one the admins (TheDrop.) came up and written it and I left it completely untouched, so it's better if you discussing it with him

kyrie25:
@Vukkyy 
Also that line is to cut off every button/text that appears _after_ that `[edit source]` button. So it should work for every other button that appears after `[edit source]`

EncryptedDev (approved):

PR #5172, opened 2021-12-28
feat(Messenger): add settings & patch presence

kyrie25:
**Add presence settings & display recipient name accordingly:**

<details>
<summary> Proof of the modification working: </summary>
<br>

image
image
image
image
image

</details>


kyrie25:
**Patch `Writing to` presence detection**
image

websites/M/Messenger/presence.ts
```diff
@@ -2,58 +2,52 @@ const presence = new Presence({
...
 presence.on("UpdateData", async () => {
   const presenceData: PresenceData = {
-    startTimestamp: browsingTimestamp
-  };
+      largeImageKey: "messenger",
+      startTimestamp: browsingTimestamp
+    },
+    [messageRecipient, callRecipient] = await Promise.all([
+      presence.getSetting("message"),
+      presence.getSetting("call")
```
Bas950, line 13: add the type

Bas950 (dismissed):

websites/M/Messenger/presence.ts
kyrie25: done

websites/M/Messenger/presence.ts
```diff
@@ -2,58 +2,52 @@ const presence = new Presence({
...
     } else {
       presenceData.details = "Writing to:";
       presenceData.smallImageKey = "writing";
     }
-    //presenceData.state = user.textContent;
-    presenceData.state = "(Hidden until presence settings.)"; // Add setting for this when presence settings are a thing!!
+    presenceData.state = messageRecipient
+      ? document.querySelector("div.t6p9ggj4.tkr6xdv7 span > span")
+          ?.textContent ?? "Unknown"
+      : "(Hidden)";
```
EncryptedDev, line 45: pref: Use an if statement here instead and do not include the `(Hidden)` part

EncryptedDev (dismissed):

websites/M/Messenger/presence.ts
kyrie25: Is it a requirement? Since if I omit the hidden part I would have to drag `.details` out on several more lines just to make it fit context (at least with my solution)
image
image

EncryptedDev: It is not a requirement, but I think it would look a bit nicer. You could set `details` accordingly too

Your call I guess

kyrie25: It looks okay now I guess, and I'll be down in a few minutes and won't be able to write so it's fine to leave it so ig.

Slowlife01 (approved):

EncryptedDev (approved):

PR #5183, opened 2021-12-29
chore: turn eslint warnings into errors

ImRodry:
This PR changes the eslint config to error on pretty much all rules except two. This is because we now have a ci that will pass even when these errors are thrown, which we do not want to happen.

<details>
<summary> Proof to proving the creation/modification working </summary>
<br>

</details>


EncryptedDev (changes requested):
Remove `no-restricted-syntax` array

EncryptedDev (approved):

Slowlife01 (approved):

EncryptedDev (dismissed):

EncryptedDev (approved):

.eslintrc.json
```diff
@@ -20,37 +20,30 @@
...
+    "comma-dangle": ["error", "never"],
+    "one-var": ["error", "consecutive"],
+    "linebreak-style": ["error", "unix"],
     "prefer-const": "error",
     "no-multi-spaces": "error",
     "object-shorthand": "error",
     "quote-props": ["error", "as-needed"],
-    "prefer-destructuring": "warn",
-    "prefer-template": "warn",
+    "prefer-destructuring": ["error", { "array": false, "object": true }],
```
Bas950, line 35: why not array?

Bas950 (changes requested):

.eslintrc.json
ImRodry: It reduces readability when working with larger indexes and it's just annoying

Bas950: Cant we have warn for array, and error for object? Cause for like `[first, second] = []` is nice for `[, , , , idk] = []` not so

ImRodry: Warnings are supposed to be fixed the same way as errors, they're just meant to be less obtrusive, so no I don't think so because that will clutter the repo and every PR going forward

Bas950 (approved):

PR #5196, opened 2021-12-30
feat(YouTube Music): Added volume setting & value

Qwerty1Verified:
redacted


down-bad:
Thanks for your contribution but, in my opinion, that's not really something that belongs to a presence.

EncryptedDev:
closed - we have no intention or reason to have this on a presence

PR #5218, opened 2022-01-01
fix(Google Keep)

down-bad:
Fix presence not working at all, added settings, updated logo

image


websites/G/Google Keep/presence.ts
```diff
@@ -9,50 +9,60 @@ presence.on("UpdateData", async () => {
       startTimestamp: timer
     },
     { hash } = location,
-    [, , , , userInt, , LorLNorId] = location.href.split("/");
+    [, , , , LorLNorId] = location.href.split("/"),
```
EncryptedDev, line 12: It is silly to use array destructuring here

EncryptedDev (changes requested):

down-bad:
Since the original dev is not responding and I want to update the logo, should I change the client id or use an image url?

EncryptedDev:
> Since the original dev is not responding and I want to update the logo, should I change the client id or use an image url?

I've seen your attempts contacting the presence developer in many different ways - feel free to add yourself as presence author & create a new `clientId`

EncryptedDev (approved):

Slowlife01 (approved):

PR #5225, opened 2022-01-02
feat(Pandora): add album art

FireController1847:
Added album art to the Pandora Rich Presence. Simple as that :)

Proof of successful modification:

https://user-images.githubusercontent.com/16494272/147870997-52e3e639-f347-413f-8a5d-f6a944ce7a3a.mp4

### DIFF
--- a/websites/P/Pandora/dist/metadata.json
+++ b/websites/P/Pandora/dist/metadata.json
@@ -21,7 +21,7 @@
     "nl": "Speel de nummers, albums, afspeellijsten en podcasts af die je leuk vindt op de geheel nieuwe Pandora."
   },
   "url": "www.pandora.com",
-  "version": "2.0.3",
+  "version": "2.1.0",
   "logo": "https://i.imgur.com/Tt7iIYq.png",
   "thumbnail": "https://i.imgur.com/fRQo8Vc.png",
   "color": "#224099",
--- a/websites/P/Pandora/dist/metadata.json
+++ b/websites/P/Pandora/dist/metadata.json
@@ -21,7 +21,7 @@
     "nl": "Speel de nummers, albums, afspeellijsten en podcasts af die je leuk vindt op de geheel nieuwe Pandora."
   },
   "url": "www.pandora.com",
-  "version": "2.0.2",
+  "version": "2.0.3",
   "logo": "https://i.imgur.com/Tt7iIYq.png",
   "thumbnail": "https://i.imgur.com/fRQo8Vc.png",
   "color": "#224099",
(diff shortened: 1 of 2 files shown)


websites/P/Pandora/dist/metadata.json
```diff
@@ -21,7 +21,7 @@
     "nl": "Speel de nummers, albums, afspeellijsten en podcasts af die je leuk vindt op de geheel nieuwe Pandora."
   },
   "url": "www.pandora.com",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
Slowlife01, line 24:
```suggestion
  "version": "2.1.0",
```

websites/P/Pandora/presence.ts
```diff
@@ -86,6 +86,27 @@ presence.on("UpdateData", async () => {
...
+      }
+
+      // If the art still does not exist, try again but this time with the hero image
+      if (!art)
+        art = document.querySelector<HTMLSourceElement>(".HeroCard__image");
+
+      // If the art exists, set the big image key to that
+      // Otherwise, use the default 'pandora' art
+      if (art) presenceData.largeImageKey = art.src;
+      else presence.error("Art was null!");
```
Slowlife01, line 109:
```suggestion
      else presence.error("Art is null!");
```

websites/P/Pandora/dist/metadata.json
FireController1847: I was thinking about this and was like, yeah this seems like a minor less bugfix change haha

EncryptedDev (changes requested):
These comments aren't really needed as they state the obvious

FireController1847:
> These comments aren't really needed as they state the obvious

I figure there's no harm in being explicit, no?

websites/P/Pandora/presence.ts
```diff
@@ -86,6 +86,27 @@ presence.on("UpdateData", async () => {
...
+      // If the art does not exist, try again but this time with the now playing info
+      if (!art) {
+        art = document.querySelector<HTMLSourceElement>(
+          ".nowPlayingTopInfo__artContainer__art :first-child :first-child"
+        );
+      }
+
+      // If the art still does not exist, try again but this time with the hero image
+      if (!art)
+        art = document.querySelector<HTMLSourceElement>(".HeroCard__image");
```
ImRodry, line 104: Use logical assignment operators or nullish coalescing
Also remove the useless comments because we can all read javascript

FireController1847: Man, you guys really like your code to be hard to understand. Alright, I'll make the changes, but I disagree with the usage of nullish coalescing/logical assignment, I think it makes far more sense using the multiple if statements.

EncryptedDev (commented):
Wouldn't the type be `HTMLImageElement` instead of `HTMLSourceElement`?

FireController1847:
> 

That is an exceptional observation. I normally work in Java so I had presumed that the HTMLSourceElement was the sub-class of HTMLImageElement but doing some research I saw that that referred to something entirely different.

Slowlife01 (approved):

EncryptedDev (approved):

PR #5622, opened 2022-01-19
chore(Deps): bump mongodb to v4

ImRodry:

This PR *finally* bumps MongoDB to the latest version and fixes the issues that come with it. I only fixed typing issues as I cannot test this but by checking the [documentation](https://github.com/mongodb/node-mongodb-native/blob/HEAD/docs/CHANGES_4.0.0.md), this should all be working as before

To explain some of the changes:

- useUnifiedTopology was removed as the behavior previously changed by setting it to true is now the default
- client.isConnected() was removed and calling client.connect() while it is already connect is safe and returns in a "no-op" (term by MongoDB themselves). Also extendedRun is only true when the client connection is closed, so that check was obsolete anyways. The check is still present through client.topology.isConnected(), however this is not documented nor typed.


EncryptedDev (approved):

Bas950 (approved):

EncryptedDev:
@Timeraa @TheDropX

Timeraa (approved):

PR #5623, opened 2022-01-19
chore: enforce the use of tabs

EncryptedDev:
@ImRodry messed this up so I fixed it


@types/premid/index.d.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ImRodry, line 1: we should probably not do this and disable it per line

ImRodry (changes requested):

Bas950 (approved):

Slowlife01 (approved):

PR #5645, opened 2022-01-23
chore: run SE instead of lint on pre-commit

EncryptedDev:
- Add styling to SE to make it aligned with debug on the SE script
- Use `SE` instead of `lint` on pre-commit to prevent unnecessary modifications & syntax enforcement
- Remove `cron` schedule from SE as there ideally shouldn't be any errors pushed with the change of pre-commit, this can still be manually triggered using the `workflow_dispatch`
- Run Syntax Enforcer to enforce styling on the repo

### DIFF
--- a/util/tools/auto/syntaxEnforcer.ts
+++ b/util/tools/auto/syntaxEnforcer.ts
@@ -9,7 +9,7 @@ import debug from "debug";
 const log = debug("SyntaxEnforcer");
 debug.enable("SyntaxEnforcer*");
 /**
- * Executes a shell command and return it as a Promise.
+ * Executes a shell command
  * @param cmd {string[]}
  * @return Promise<string>
  */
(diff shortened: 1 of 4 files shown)


util/tools/auto/syntaxEnforcer.ts
```diff
@@ -82,7 +83,7 @@ const readFile = (path: string): string =>
 	main = async (): Promise<void> => {
 		if (!process.env.GITHUB_ACTIONS) {
 			console.log(
-				"\nPlease note that this script is ONLY supposed to run on a CI environment\n"
+				chalk.red("\nThis script is ONLY supposed to run on a CI environment\n")
```
ImRodry, line 86: This is no longer true, just remove it

EncryptedDev:
@Bas950 @Slowlife01

Slowlife01 (commented):
wouldn't the version be bumped 2 times then if it was already bumped

EncryptedDev:
> wouldn't the version be bumped 2 times then if it was already bumped

What?

Bas950:
> > wouldn't the version be bumped 2 times then if it was already bumped
> 
> What?

Yea doesn't the script bump versions? Or does semid do that manually??

PR #5665, opened 2022-01-25
fix(UpBeatRadio): change query selector to get correct data

jqcktalks:
The presence was showing incorrect data when viewing news articles previously, to combat this, we added some code on the website to make sure that the presence can get the correct data, by modifying the code on our articles page. I've edited the presence code accordingly, to make this show up properly on user's Discord profiles.


Slowlife01 (changes requested):
Bump version

PR #5671, opened 2022-01-26
chore: improve SE and fix commit hooks

ImRodry:

This PR *actually* improves SE and fixes the commit hook
Closes #5645


package.json
```diff
@@ -5,9 +5,11 @@
 	"license": "MPL-2.0",
 	"scripts": {
 		"lint": "prettier --write . && eslint --fix .",
-		"lint:ci": "eslint .",
+		"lint:ci": "prettier --check . && eslint .",
 		"inc": "tsc -p ./util/configs/tsconfig.massVer.json && node ./util/tools/massVer.js",
-		"se": "tsc -p ./util/configs/tsconfig.auto.syntaxEnforcer.json && node ./util/tools/auto/syntaxEnforcer.js",
+		"incChanged": "tsc -p ./util/configs/tsconfig.auto.syntaxEnforcer.json && node ./util/tools/bumpChanged",
```
Bas950, line 10:
```suggestion
		"bumpChanged": "tsc -p ./util/configs/tsconfig.auto.syntaxEnforcer.json && node ./util/tools/bumpChanged",
```

Bas950 (changes requested):
change it to match file name

package.json
ImRodry: No bc consistency

EncryptedDev:
You changed all those metadata's but never bumped the version?

package.json
Bas950: I would either change the file name or the package script name. So they are the same like all other scripts.

ImRodry: For the same reason that massVer is called inc in the script, this is called incChanged as it does the same thing to the files in the diff.

Bas950: Well while you are at it you can make the scripts and file names consistent

ImRodry: That’s unnecessary, inc and incChanged are much easier to understand that two letter acronyms

ImRodry:
Realized it's not possible to use prettier/eslint on json files since bumping the versions would undo the changes with JSON.stringify. I added *.json to the prettier and eslint ignore files and only formatted 1 file. This is ready to be merged now.

Bas950:
> Realized it's not possible to use prettier/eslint on json files since bumping the versions would undo the changes with JSON.stringify. I added *.json to the prettier and eslint ignore files and only formatted 1 file. This is ready to be merged now.

cant you after bumping only run it on json files?

ImRodry:
You’d need to be running it twice, there’s no point and this format is just fine

Slowlife01 (approved):

Bas950 (approved):

Issue #5732, opened 2022-02-05
Add domain or url


arthillya:
### Presence name

HBO go

### Description

Can you please add this as one of the supported url for HBO go so it will show up as a status on my discord profile. Thank youuu.

This is the link

hbogoasia.ph

### Alternatives

Add supported URL hbogoasia.ph

Kaisarion:
I have submitted pull request #5739 which will add support for this URL. So sorry for the hassle

PR #5736, opened 2022-02-05
update hbogo sites to include philippines url

Kaisarion:
Contributing from the issue request [here](https://github.com/PreMiD/Presences/issues/5732).

I do not live in the Philippines so I cannot test this but I am 100% sure it does work due to the link, etc.


Kaisarion:
need to fix apple issue first. will revisit this

PR #5795, opened 2022-02-14
feat(HBO Max): cover art

### LINKED ISSUE #5596: HBO Max - Add thumbnail options
### Presence name

HBO Max

### Description

The option to view the thumbnails for shows' and movies' covers in the large image of the presence, rather than just the HBO Max logo. 

### Alternatives

_No response_

### Additional context

An example of this working would be Netflix which does the same, or even YouTube which adds the option for the channel's image or the thumbnail for the video itself (although that doesn't apply here, the way Netflix works would be just fine).

Slowlife01:
Resolves #5596

<details>
<br>

image

</details>

### DIFF
--- a/websites/H/HBO Max/presence.ts
+++ b/websites/H/HBO Max/presence.ts
@@ -44,13 +44,32 @@ const presence = new Presence({
 		"sesame-workshop": "Sesame Workshop",
 		"looney-tunes": "Looney Tunes",
 		crunchyroll: "Crunchyroll Collection"
-	};
+	},
+	coverUrls: Record<string, string> = {};
+
+function fetchSeries(): Promise<string> {
+	return new Promise(resolve => {
+		fetch(
+			`https://comet.api.hbo.com/express-content/${
+				location.pathname.split("/")[2]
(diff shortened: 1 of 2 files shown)


websites/H/HBO Max/presence.ts
```diff
@@ -44,13 +44,32 @@ const presence = new Presence({
...
+			);
+	});
+}
 
 presence.on("UpdateData", async () => {
 	const presenceData: PresenceData = {
 			largeImageKey: "lg"
 		},
-		video: HTMLVideoElement = document.querySelector("video"),
+		video = document.querySelector("video"),
```
EncryptedDev, line 72: Why did you remove the type?
```suggestion
		video = document.querySelector<HTMLVideoElement>("video"),
```

EncryptedDev (changes requested):

websites/H/HBO Max/presence.ts
Slowlife01: Because it's unnecessary, this query already gave you a type...

EncryptedDev (approved):

Bas950 (approved):

PR #6197, opened 2022-05-11
feat(Google): add privacy mode

### LINKED ISSUE #6118: Google | Privacy Mode
### Presence name

Google

### Description

add privacy options to google rpc. to hide searches

### Alternatives

_No response_

### Additional context

_No response_

kyrie25:
Resolves #6118 
<details>
<br>

image
image

</details>

### DIFF
--- a/websites/G/Google/presence.ts
+++ b/websites/G/Google/presence.ts
@@ -14,12 +14,9 @@ presence.on("UpdateData", async () => {
 		},
 		privacy = await presence.getSetting<boolean>("privacy");
 
-	if ((homepageInput && homepageImage) || !document.location.pathname) {
-		presenceData.state = "Home";
-		presenceData.startTimestamp = browsingTimestamp;
-
-		delete presenceData.details;
-	} else if (document.location.pathname.startsWith("/doodles/")) {
+	if ((homepageInput && homepageImage) || !document.location.pathname)
+		presenceData.details = "Home";
+	else if (document.location.pathname.startsWith("/doodles/")) {
 		const doodleResult = new URL(document.location.href).searchParams.get("q"),
 			doodleTitle: HTMLElement = document.querySelector(
 				"#title-card > div > h2"
--- a/websites/G/Google/presence.ts
+++ b/websites/G/Google/presence.ts
@@ -9,9 +9,10 @@ const presence = new Presence({
 
 presence.on("UpdateData", async () => {
 	const presenceData: PresenceData = {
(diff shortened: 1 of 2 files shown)


Slowlife01 (commented):
See line 32 & 39

kyrie25:
> See line 32 & 39

My `if (privacy)` statement is contained within `else if (document.location.pathname.startsWith("/search"))` so it won't delete the `state` strings of the lines you mentioned, if that is what you're talking about?

websites/G/Google/presence.ts
```diff
@@ -9,9 +9,10 @@ const presence = new Presence({
...
-		largeImageKey: "lg",
-		startTimestamp: browsingTimestamp
-	};
+			largeImageKey: "lg",
+			startTimestamp: browsingTimestamp
+		},
+		privacy = await presence.getSetting<boolean>("privacy");
 
 	if ((homepageInput && homepageImage) || !document.location.pathname) {
 		presenceData.state = "Home";
```
ImRodry, line 18: no clue why this is in the state but I think you can mention the user is in home with privacy on, so this should go to the details. Also do me a favour and remove the line right below this one as it's unnecessary

websites/G/Google/presence.ts
```diff
@@ -86,7 +87,11 @@ presence.on("UpdateData", async () => {
 				// No default
 			}
 		}
+		if (privacy) {
+			delete presenceData.state;
+			if (presenceData.details.includes("Searching for"))
```
ImRodry, line 92: Due to line 32, this will say "Viewing a Doodle:" with nothing below, so you should probably make another if statement to account for that

ImRodry (changes requested):

websites/G/Google/presence.ts
kyrie25: I specifically just said on my comment above that my `if (privacy)` statement is contained and only affects while user is on `/search` lol. It does not effect the doodle case

kyrie25: And `Viewing Doodle` is not privacy-invasive so I don't think you need one for that

websites/G/Google/presence.ts
kyrie25: I'll do it ASAP

ImRodry (approved):
Thank you for this PR!

Slowlife01 (approved):

PR #6203, opened 2022-05-12
refactor(AniCloud): overhaul AniCloud presence because of rebrand to AniWorld

### LINKED ISSUE #6146: Change AniCloud Name and Url to match rebrand
### Presence name

AniCloud

### Description

AniCloud.io rebranded to AniWorld.to.
The PreMiD-Presence that was used on AniCloud.io is now not working, since it is a new url.

Fixing this should be pretty simple since its a small rename.

### Alternatives

_No response_

### Additional context

AniCloud.io redirects to AniWorld.to now.

Name-Change Information can be found on their website [here](https://aniworld.to/support/frage/neu-aniworld-alt-anicloud)

PlayWolfYT:
refactor(AniCloud): rebrand to AniWorld

Change URL from "AniCloud.io" to "AniWorld.to", because there was a rebrand.

close #6146
BREAKING CHANGE: url is completely different, therefore the old presence does not work at all at the moment


websites/A/AniWorld/presence.ts
```diff
@@ -27,15 +23,23 @@ const presence = new Presence({
...
 	currentTime: number,
 	paused: boolean,
 	played: boolean;
 
+async function getStrings() {
+	return presence.getStrings({
+		play: "presence.playback.playing",
+		pause: "presence.playback.paused"
+	});
+}
```
Slowlife01, line 41:
```suggestion
function getStrings() {
	return presence.getStrings({
		play: "general.playing",
		pause: "general.paused"
	});
}
```

Slowlife01 (dismissed):

websites/A/AniWorld/dist/metadata.json
```diff
@@ -4,23 +4,24 @@
...
+		"vi_VN": "Xem ngay hơn 1,000+ bộ anime tại AniWorld hoàn toàn miễn phí. Anime theo nhu cầu: ✓ 100% miễn phí ✓ Trực tuyến ✓ Những bộ anime độc quyền"
 	},
-	"url": "anicloud.io",
-	"version": "1.0.13",
-	"logo": "https://i.imgur.com/eCZpN6e.png",
-	"thumbnail": "https://i.imgur.com/fKBCijM.png",
+	"url": "aniworld.to",
+	"version": "1.0.0",
+	"logo": "https://i.imgur.com/fM3E5Iv.png",
+	"thumbnail": "https://i.imgur.com/EPM1zOf.png",
```
ImRodry, line 17: could you try to get a wider image? The closest to 16:9 the better

websites/A/AniWorld/presence.ts
```diff
@@ -49,73 +53,75 @@ presence.on(
...
+			}
+			if (played) {
+				if (!paused) {
+					[, presenceData.endTimestamp] = presence.getTimestamps(
+						currentTime,
+						timeEnd
+					);
+				}
+				presenceData.smallImageKey = paused ? "pause" : "play";
+				presenceData.smallImageText = paused ? strings.pause : strings.play;
```
ImRodry, line 102: This PR is meant to be a rebrand not a refactor from what I understand. If you want to completely overhaul the presence then you should probably change the title of the PR. Otherwise I would recommend undoing these

ImRodry (changes requested):

websites/A/AniWorld/dist/metadata.json
PlayWolfYT: Thumbnail updated

websites/A/AniWorld/presence.ts
PlayWolfYT: I changed the PR title. Not sure how to describe it, since english is not my main language, hope this is okay.

ImRodry: so did the entire website change or did only the URL change?

PlayWolfYT: only the url as far i'm aware, but during testing I noticed a few things that were broken (based on urls that existed previously that are no longer existing). Those we're fixed a few commits ago

websites/A/AniWorld/presence.ts
```diff
@@ -49,73 +53,75 @@ presence.on(
 );
 
 presence.on("UpdateData", async () => {
+	if (!strings) strings = await getStrings();
```
ImRodry, line 56: You shouldn't use getStrings if you don't plan on supporting multiLanguage. Either remove this and use static strings or add proper support for multiLanguage for all strings (I recommend the first option)

ImRodry (changes requested):

websites/A/AniWorld/presence.ts
PlayWolfYT: Strings removed.

ImRodry (approved):

Miyakura (changes requested):
"Exklusive Animes" - otherwise the translation is correct
I would personally simplify "Wiedergabe pausiert" with "Pause" or "Pausiert".

EncryptedDev (approved):

PR #6171, opened 2022-05-04
feat(Wolfy): add multiLanguage support & handle event feature

Tenebrosful:
**Describe the changes in this pull request:**

- All strings are now dynamic and change with Wolfy's lang
- Handling Event new Wolfy feature

<details>
<summary> Proof showing the creation/modification is working as expected </summary>
<br>

image

image

image

</details>

PR #6224, opened 2022-05-16
fix(Wolfy): lang path offset and handle events

Tenebrosful:

- Fix the path usage caused by the addition of english games
- Handle events

Originally from #6171

### DIFF
--- a/websites/W/Wolfy/presence.ts
+++ b/websites/W/Wolfy/presence.ts
@@ -9,6 +9,11 @@ let path,
 	cp: number,
 	currTime: string;
 
+const waitingString: Record<string, string> = {
+	en: "WAITING",
+	fr: "EN ATTENTE"
+};
+
 function getTime(list: string[]): number {
 	let ret = 0;
 	for (let index = list.length - 1; index >= 0; index--)
@@ -48,6 +53,10 @@ function addConsultCategoryButton(presenceData: PresenceData, url: string) {
 	addButton(presenceData, "Consulter la catégorie", url);
 }
 
+function addConsultEventButton(presenceData: PresenceData, href: string) {
+	addButton(presenceData, "Consulter l'évènement", href);
+}
+
 async function addJoinGameButton(presenceData: PresenceData, gameId: string) {
 	if (!(await presence.getSetting("joinGameButton"))) return;
 	addButton(
@@ -55,7 +64,7 @@ async function addJoinGameButton(presenceData: PresenceData, gameId: string) {
 		`(${
 			document.querySelector("div.Header_timer__36MsP")?.textContent
 		}) Rejoindre la partie`,
-		`https://wolfy.fr/game/${gameId}`
+		`https://wolfy.net/game/${gameId}`
 	);
 }
 
@@ -63,7 +72,7 @@ function addVisitHelpCenterButton(presenceData: PresenceData) {
 	addButton(
 		presenceData,
 		"Consulter le centre d'aide",
-		"https://help.wolfy.fr"
+		"https://help.wolfy.net"
 	);
 }
 
@@ -77,11 +86,11 @@ async function addVisitProfilButton(
 	let label = `Visiter le profil de ${username}`;
 	if (label.length > 32) label = `${label.slice(0, 31)}…`;
 
-	addButton(presenceData, label, `https://wolfy.fr/leaderboard/${username}`);
+	addButton(presenceData, label, `https://wolfy.net/leaderboard/${username}`);
 }
 
 function addVisitWolfyButton(presenceData: PresenceData) {
-	addButton(presenceData, "Visiter le site de Wolfy", "https://wolfy.fr");
+	addButton(presenceData, "Visiter le site de Wolfy", "https://wolfy.net");
 }
 
 async function handleCheckingLeaderboard(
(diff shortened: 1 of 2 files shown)


websites/W/Wolfy/presence.ts
```diff
@@ -9,6 +9,11 @@ let path,
 	cp: number,
 	currTime: string;
 
+const waitingString: Record<string, string> = {
```
ImRodry, line 12:
```suggestion
const waitingString = {
```

websites/W/Wolfy/presence.ts
```diff
@@ -178,35 +190,59 @@ presence.on("UpdateData", async () => {
 			currTime = document.querySelector("div.Header_timer__36MsP").textContent;
 		}
 
+		presenceData.details = "En jeu";
+
+		presenceData.smallImageKey = "live";
+		presenceData.smallImageText = "En jeu";
+		if (currTime?.includes(":")) {
+			[presenceData.startTimestamp, presenceData.endTimestamp] = getTimestamps(
```
ImRodry, line 198: you could probably use presence.getTimestampsFromFormat here but I'm not entirely sure, you should test that

ImRodry (changes requested):

itsmelouis (approved):
Looks good to me about the translations

websites/W/Wolfy/presence.ts
Tenebrosful: The program don't compile without this type specefication because it's used like an array with string keys like you can see line 210

```ts
presenceData.state === waitingString[document.querySelector("html").lang]
```

websites/W/Wolfy/presence.ts
Tenebrosful: It was from the original code made by @fr0r, i didn't really looked deeper about this and didn't changed anything

websites/W/Wolfy/presence.ts
ImRodry: then you have to check if the query selector is either "en" or "fr" for type safety, this is not the solution

websites/W/Wolfy/presence.ts
ImRodry: well then why is this in the diff?

Tenebrosful: Just moved it before the handling of the state "WAITING" in game to set image and time before handling specific game's state, there is no impact on running

websites/W/Wolfy/presence.ts
Tenebrosful: I did in my last commit 👍

websites/W/Wolfy/presence.ts
```diff
@@ -160,82 +182,104 @@ presence.on("UpdateData", async () => {
...
 			prevState = presenceData.state;
 			cp = Date.now();
-			currTime = document.querySelector("div.Header_timer__36MsP").textContent;
+			currTime = document.querySelector("div.Header_timer__xtmy2").textContent;
 		}
 
+		presenceData.details = "En jeu";
+
+		presenceData.smallImageKey = "live";
+		presenceData.smallImageText = "En jeu";
```
ImRodry, line 206: the details say the exact same thing as this so you can just remove this line

websites/W/Wolfy/presence.ts
ImRodry: then why did you move it?

websites/W/Wolfy/presence.ts
```diff
@@ -93,45 +102,58 @@ async function handleCheckingLeaderboard(
...
+
 presence.on("UpdateData", async () => {
 	const presenceData: PresenceData = {
 		largeImageKey: "logo"
 	};
 
 	path = document.location.pathname;
 
+	const pathOffset =
+		path.split("/")[1] === document.querySelector("html")?.lang ? 1 : 0; // If the language is the first path element, we need to add an offset
```
ImRodry, line 155: instead of doing this, you should make a constant in the event for `path.split("/")` to avoid computing it time and time again and pop out the element you need to pop out, I believe that's more efficient

ImRodry (changes requested):

websites/W/Wolfy/presence.ts
Tenebrosful: Just for code visibility to assign general things before specefic things.

Can revert it if you want

websites/W/Wolfy/presence.ts
Tenebrosful: Idk if this was what you was talking about but i just did it in commit [c110352](https://github.com/PreMiD/Presences/pull/6224/commits/c1103523cb526056aab767695a51997f623351f6)

websites/W/Wolfy/presence.ts
ImRodry: If this has no practical impact please do

Tenebrosful: Revert done

ImRodry (approved):
Make sure it all still works with the latest changes, otherwise LGTM

Slowlife01 (approved):
Might want to try https://www.w3schools.com/cssref/sel_attr_begin.asp if the classes' name get changed again

Tenebrosful:
> Might want to try https://www.w3schools.com/cssref/sel_attr_begin.asp if the classes' name get changed again

Yeah i will take a look at but i talked to dev and they will maybe change all class to fully random generate to make Bot protection :/

PR #6311, opened 2022-06-11
feat(Crowdin): support website redesign and general improvement

ImRodry:
## Description 

The Crowdin website recently received a major update that completely changed the look of every page so I fixed all of them and added support for missing pages. This update also comes with two new settings:

- Show project names: toggles whether the name of the current project will be displayed on the presence
- Show manager pages: toggles whether manager-specific pages will be shown. When disabled it will show as if the user was on the main page
- Show conversations: toggles whether message subject/participants are shown. When disabled the presence will only say whether the person is messaging a user or a group.

I attached a lot of screenshots for proof. If the functionality of a certain setting or a page is unclear I can provide more screenshots.

## Acknowledgements
- [X] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
- [X] I linted the code by running `yarn format`
- [X] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>

image
image
image
image
image
image
image
image
image
image
image
image
image
Show messages disabled
image
Show messages enabled
image

</details>

### DIFF
--- a/websites/C/Crowdin/dist/metadata.json
+++ b/websites/C/Crowdin/dist/metadata.json
@@ -16,7 +16,8 @@
 		"ro": "Crowdin este o soluție bazată pe cloud care eficientizează gestionarea localizării pentru echipa ta.",
 		"nl": "Crowdin is een cloudgebaseerde oplossing die lokalisatiebeheer voor uw team stroomlijnt.",
 		"vi_VN": "Crowdin là giải pháp trực tuyến cho việc quản lý đội ngũ bản địa hoá của bạn.",
-		"pt": "O Crowdin é uma solução de cloud que facilita a gestão de traduções para a tua equipa."
+		"pt": "O Crowdin é uma solução de cloud que facilita a gestão de traduções para a tua equipa.",
+		"ru": "Crowdin — это облачное решение, которое упрощает управление локализацией для вашей команды."
 	},
 	"url": [
 		"crowdin.com",
--- a/websites/C/Crowdin/dist/metadata.json
+++ b/websites/C/Crowdin/dist/metadata.json
@@ -39,6 +39,12 @@
 	],
 	"settings": [
 		{
+			"id": "showProject",
+			"title": "Show project names",
+			"icon": "fa-regular fa-tag",
+			"value": true
+		},
+		{
 			"id": "showManager",
 			"title": "Show manager pages",
 			"icon": "fa-duotone fa-user-police-tie",
--- a/websites/C/Crowdin/dist/metadata.json
+++ b/websites/C/Crowdin/dist/metadata.json
@@ -15,17 +15,17 @@
 		"en": "Crowdin is a cloud-based solution that streamlines localization management for your team.",
 		"ro": "Crowdin este o soluție bazată pe cloud care eficientizează gestionarea localizării pentru echipa ta.",
 		"nl": "Crowdin is een cloudgebaseerde oplossing die lokalisatiebeheer voor uw team stroomlijnt.",
-		"vi_VN": "Crowdin là giải pháp trực tuyến cho việc quản lý đội ngũ bản địa hoá của bạn."
+		"vi_VN": "Crowdin là giải pháp trực tuyến cho việc quản lý đội ngũ bản địa hoá của bạn.",
+		"pt": "O Crowdin é uma solução de cloud que facilita a gestão de traduções para a tua equipa."
 	},
 	"url": [
-		"translate.premid.app",
 		"crowdin.com",
 		"support.crowdin.com",
 		"store.crowdin.com",
 		"status.crowdin.com",
 		"blog.crowdin.com"
 	],
-	"version": "3.1.2",
+	"version": "4.0.0",
 	"logo": "https://i.imgur.com/rtkC4kS.png",
 	"thumbnail": "https://i.imgur.com/dJnnzqd.png",
 	"color": "#2e3340",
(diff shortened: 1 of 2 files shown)


websites/C/Crowdin/presence.ts
```diff
@@ -3,96 +3,124 @@ const presence = new Presence({
...
-						? activeLabel?.children[1].textContent
-							? activeLabel?.children[1].textContent
-							: activeLabel?.children[0].textContent
-						: activeLabel?.textContent;
+
+				presenceData.details = `Reading the ${
+					isEnterprise ? "Enterprise " : ""
+				}API documentation`;
+
+				presenceData.state = [...(activeLabel?.children ?? [])]
```
Bas950, line 54: so if its empty dus it return null/undefined or empty string

ImRodry: it maps an empty array and joins is thus making it an empty string

Bas950: do we allow empty strings as state? I thought we said you must just delete presencedata.state then

ImRodry: Deleting and having an empty string produces the same result in the end. Doing it like this is more practical, what we usually don't allow is explicitly setting the state or any other property to an empty string. If it's just a possible outcome of a larger code block I think it's fine

ImRodry:
Marking as draft temporarily while I add a setting to toggle showing the project name

websites/C/Crowdin/presence.ts
EncryptedDev: > do we allow empty strings as state? I thought we said you must just delete presencedata.state then

This is correct

ImRodry: there's no point in deleting here since I'm not explicitly setting it to an empty string. The end result is the same, performance wise it would be worse to delete it

websites/C/Crowdin/dist/metadata.json
```diff
@@ -15,24 +15,40 @@
 		"en": "Crowdin is a cloud-based solution that streamlines localization management for your team.",
 		"ro": "Crowdin este o soluție bazată pe cloud care eficientizează gestionarea localizării pentru echipa ta.",
 		"nl": "Crowdin is een cloudgebaseerde oplossing die lokalisatiebeheer voor uw team stroomlijnt.",
-		"vi_VN": "Crowdin là giải pháp trực tuyến cho việc quản lý đội ngũ bản địa hoá của bạn."
+		"vi_VN": "Crowdin là giải pháp trực tuyến cho việc quản lý đội ngũ bản địa hoá của bạn.",
+		"pt": "O Crowdin é uma solução de cloud que facilita a gestão de traduções para a tua equipa."
```
Syjalo, line 19:
```suggestion
		"pt": "O Crowdin é uma solução de cloud que facilita a gestão de traduções para a tua equipa.",
		"ru": "Crowdin — это облачное решение, которое упрощает управление локализацией для вашей команды."
```

Syjalo (changes requested):

ImRodry:
Feature added and ready for review

websites/C/Crowdin/dist/metadata.json
```diff
@@ -15,24 +15,47 @@
...
+		"pt": "O Crowdin é uma solução de cloud que facilita a gestão de traduções para a tua equipa.",
+		"ru": "Crowdin — это облачное решение, которое упрощает управление локализацией для вашей команды."
 	},
 	"url": [
-		"translate.premid.app",
 		"crowdin.com",
 		"support.crowdin.com",
 		"store.crowdin.com",
 		"status.crowdin.com",
 		"blog.crowdin.com"
```
Syjalo, line 27: What about links with language codes?
https://ru.crowdin.com/
https://pt.crowdin.com/

Syjalo (changes requested):

websites/C/Crowdin/dist/metadata.json
ImRodry: Could've sworn we had a regex in place, will update

EncryptedDev (approved):
LGTM once you commit

websites/C/Crowdin/dist/metadata.json
```diff
@@ -15,24 +15,48 @@
...
 	],
-	"version": "3.1.2",
+	"regExp": "[a-z]{2}\\.crowdin\\.com",
```
Slowlife01, line 29: `support.crowdin.com` and other urls won't work with this regex, the urls in the array are not used as fallback

Slowlife01 (changes requested):

websites/C/Crowdin/dist/metadata.json
ImRodry: This is intentional. The regex is not meant to replace the urls array

Slowlife01: > the urls in the array are not used as fallback

You should see how it works
https://github.com/PreMiD/Extension/blob/release/src/util/tabPriority.ts#L66-L73

ImRodry: Now that’s stupid
In that case I won’t support translated Crowdin domains for now

ImRodry: removed

Slowlife01 (approved):

PR #6321, opened 2022-06-13
fix(Coub): updated page regex, adjusted buttons array

Venipa:
## Description 
fixed error on presence.buttons.push was of undefined, updated match regex, fixed initial string fetch

## Acknowledgements

## Screenshots
image

### DIFF
--- a/websites/C/Coub/presence.ts
+++ b/websites/C/Coub/presence.ts
@@ -53,8 +53,8 @@ const pages: PageContext[] = [
 		{
 			middleware: ref =>
 				!!ref.location.pathname.match(
-					/^\/(hot|tags|rising|fresh|feed|rising|stories|random|bookmarks|likes|weekly|best|stories|royal\.coubs)/gi
-				) || ref.location.pathname === "/",
+					/^\/?(hot|tags|rising|fresh|feed|rising|stories|random|bookmarks|likes|weekly|best|stories|royal\.coubs|\/)/gi
+				),
 			exec: (
 				context,
 				data,
@@ -118,19 +118,17 @@ const pages: PageContext[] = [
 						? images.PLAY
 						: images.PAUSE;
 				if (showWatch) {
-					data.buttons.push(
-						...[
-							{
-								label: strings.watchVideo,
-								url: `${document.location.origin}/view/${activeMedia.dataset.permalink}`,
-							},
-							...getSourceLink(
-								activeMedia.querySelector<HTMLAnchorElement>(
-									".description__stamp a.description__stamp__source"
-								)?.href
-							),
-						]
-					);
+					data.buttons = [
+						{
+							label: strings.watchVideo,
+							url: `${document.location.origin}/view/${activeMedia.dataset.permalink}`,
+						},
+						getSourceLink(
+							activeMedia.querySelector<HTMLAnchorElement>(
+								".description__stamp a.description__stamp__source"
+							)?.href
+						)?.[0],
+					];
 				}
 				return data;
 			},
@@ -183,20 +181,18 @@ const pages: PageContext[] = [
 						? images.PLAY
 						: images.PAUSE;
 				if (showWatch) {
-					data.buttons.push(
-						...[
-							{
-								label: strings.watchVideo,
-								url: `${document.location.origin}/view/${activeMedia.dataset.permalink}`,
-							},
-							{
-								label: strings.viewProfil,
-								url: `${document.location.origin}/${
-									document.location.pathname.split("/")[1]
-								}`,
-							},
-						]
-					);
+					data.buttons = [
+						{
+							label: strings.watchVideo,
+							url: `${document.location.origin}/view/${activeMedia.dataset.permalink}`,
+						},
+						{
+							label: strings.viewProfil,
+							url: `${document.location.origin}/${
+								document.location.pathname.split("/")[1]
+							}`,
+						},
+					];
 				}
 				return data;
 			},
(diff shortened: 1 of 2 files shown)


ImRodry:
Rebased to apply the latest changes in the prettier configuration and fix conflicts. If you need help with your local branch, please ping me on the #presence-dev channel on Discord.

websites/C/Coub/presence.ts
```diff
@@ -312,26 +304,28 @@ const pages: PageContext[] = [
 	],
 	presence = new Presence({
 		clientId: "818598086984728576",
-	});
-
+	}),
+	IMAGES = {
```
EncryptedDev, line 308: Shouldn't this be lowercase

EncryptedDev (changes requested):

websites/C/Coub/presence.ts
Venipa: dunno, lint didnt show any warnings

ImRodry: lint doesn't warn for that but that's not the capitalization standard we follow

websites/C/Coub/presence.ts
```diff
@@ -231,19 +227,17 @@ const pages: PageContext[] = [
...
+					data.buttons = [
+						{
+							label: strings.watchVideo,
+							url: document.location.href,
+						},
+						getSourceLink(
+							activeMedia.parentElement.querySelector<HTMLAnchorElement>(
+								'.coub__info .media-block__item > a[type="embedPopup"]'
+							)?.href
+						)?.[0],
```
ImRodry, line 239: getSourceLink never returns an array with more than 1 element, why is it even returning an array to begin with?

ImRodry (changes requested):

ImRodry:
The remaining reviews on this PR must be addressed in a timely manner, otherwise this will be closed in the next few days

websites/C/Coub/presence.ts
```diff
@@ -32,29 +31,27 @@ function getQuery() {
...
+	/^((?:https?:)?\/\/)?((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w-]+\?v=|embed\/|v\/)?)([\w-]+)(\S+)?$/gm;
 function matchYoutubeUrl(url: string): boolean {
-	return !!url.match(
-		/^((?:https?:)?\/\/)?((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w-]+\?v=|embed\/|v\/)?)([\w-]+)(\S+)?$/gm
-	);
+	return youtubeUrlRegex.test(url);
 }
-function getSourceLink(url: string): { label: string; url: string }[] {
-	if (!url) return [];
+function getSourceLink(url: string): { label: string; url: string } | null {
```
ImRodry, line 39:
```suggestion
function getSourceLink(url: string): ButtonData | null {
```

websites/C/Coub/presence.ts
```diff
@@ -32,29 +31,27 @@ function getQuery() {
 function capitalizeFirstLetter(string: string) {
 	return string.charAt(0).toUpperCase() + string.slice(1);
 }
+const youtubeUrlRegex =
+	/^((?:https?:)?\/\/)?((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w-]+\?v=|embed\/|v\/)?)([\w-]+)(\S+)?$/gm;
 function matchYoutubeUrl(url: string): boolean {
-	return !!url.match(
-		/^((?:https?:)?\/\/)?((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w-]+\?v=|embed\/|v\/)?)([\w-]+)(\S+)?$/gm
-	);
+	return youtubeUrlRegex.test(url);
```
ImRodry, line 37: do you really need this function? The regex is already a constant, surely you can just call .test on it

websites/C/Coub/presence.ts
```diff
@@ -312,26 +301,28 @@ const pages: PageContext[] = [
...
+		{
+			browsing: "presence.activity.browsing",
+			watching: "presence.playback.playing",
+			watchVideo: "general.buttonWatchVideo",
+			viewProfile: "general.buttonViewProfile",
+		},
+		newLang
+	);
+}
 let currentLang: string, localizedStrings: { [key: string]: string };
```
ImRodry, line 320:
```suggestion
let currentLang: string, localizedStrings: Awaited<typeof getStrings>;
```

websites/C/Coub/presence.ts
```diff
@@ -312,26 +301,28 @@ const pages: PageContext[] = [
...
-const IMAGES = {
-	PLAY: "playx1024",
-	PAUSE: "pausex1024",
-};
 presence.on("UpdateData", async () => {
-	const newLang = await presence.getSetting<string>("lang");
-	if (newLang !== currentLang) {
+	const newLang = await presence.getSetting<string>("lang").catch(() => null);
```
ImRodry, line 322: why null? We usually catch this to "en"

ImRodry (changes requested):
Final nits

websites/C/Coub/presence.ts
Venipa: Presence.getStrings accepts null/undefined, thought it would resolve to "en" if null/undefined

websites/C/Coub/presence.ts
Venipa: doesnt work, getStrings is a function.
defined the strings in a const and ref the type to that

websites/C/Coub/presence.ts
ImRodry: That’s not how JS works, when you use .catch and the promise rejects it will resolve to the value the catch function returns

websites/C/Coub/presence.ts
ImRodry: Yes it does work, I just forgot something. Should be like this:
`Awaited<ReturnType<typeof getStrings>>`
Please change this

websites/C/Coub/presence.ts
```diff
@@ -312,26 +295,27 @@ const pages: PageContext[] = [
...
+	presenceImageKeys = {
+		PLAY: "playx1024",
+		PAUSE: "pausex1024",
+	},
+	presenceStrings = {
+		browsing: "presence.activity.browsing",
+		watching: "presence.playback.playing",
+		watchVideo: "general.buttonWatchVideo",
+		viewProfile: "general.buttonViewProfile",
+	};
```
ImRodry, line 308: This does not need to be a constant, it can just be passed onto the function

EncryptedDev: Unresolved

EncryptedDev (changes requested):

Bas950 (approved):

EncryptedDev (approved):

PR #6336, opened 2022-06-17
fix(DashNet): cookie clicker spacing issue

Cloweeee:
## Description 

There is no space between the integer and the shortened number. (e.g.: million, billion, trillion)
I have added a function that adds a space in between these values, as well as keeping all the working rules in place.
This works with both the shortened and long number format.

## Acknowledgements
- [X] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)

## Screenshots
<details>

**Before**:
A screenshot of my friend, that is using the version from the store:
image

**After**:
It works with shortened numbers:
image

As well as long numbers: 
image

</details>

### DIFF
--- a/websites/D/DashNet/dist/metadata.json
+++ b/websites/D/DashNet/dist/metadata.json
@@ -4,6 +4,12 @@
 		"name": "Bas950",
 		"id": "241278257335500811"
 	},
+	"contributors": [
+		{
+			"name": "DananaBanana",
+			"id": "423478609529929728"
+		}
+	],
 	"service": "DashNet",
 	"altnames": [
 		"Cookie Clicker"
@@ -14,7 +20,7 @@
 	},
 	"url": "dashnet.org",
 	"regExp": "([a-z0-9-]+[.])*dashnet[.]org[/]",
-	"version": "1.3.5",
+	"version": "1.3.6",
 	"logo": "https://i.imgur.com/xcK0UqX.png",
 	"thumbnail": "https://i.imgur.com/PnsZ4Nv.png",
 	"color": "#042B48",
@@ -22,5 +28,11 @@
 	"tags": [
 		"cookie",
 		"clicker"
+	],
+	"contributors": [
+		{
+			"name": "DananaBanana",
+			"id": "423478609529929728"
+		}
 	]
 }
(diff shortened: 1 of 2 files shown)


ImRodry:
Please show a before and after

DananaBanana:
Edited the post

websites/D/DashNet/presence.ts
```diff
@@ -14,6 +14,20 @@ function presenceSet(): void {
...
+	for (const c of str.split("")) {
+		newStr +=
+			isNaN(Number(c)) && found == false && c !== "." && c !== ","
+				? ` ${c}`
+				: c;
+		if (newStr.includes(" ")) found = true;
+	}
+
+	return newStr;
+}
```
ImRodry, line 29: this whole function is unnecessarily complex and is fixing an issue that shouldn't exist to begin with. In order to fix this you need to use the proper query selector when getting the text

DananaBanana: Fixed this with a new commit, replaced the old function with a more efficient query selector.

websites/D/DashNet/presence.ts
```diff
@@ -29,10 +29,9 @@ presence.on("UpdateData", () => {
 	if (document.location.pathname.includes("/cookieclicker/")) {
 		const cookies = document
 			.querySelector("#cookies")
-			.textContent.replace(
-				document.querySelector("#cookies div").textContent,
-				""
-			);
+			.innerHTML.replace("<br>", " ")
+			.split("<")[0];
```
ImRodry, line 33: You should never use innerHTML because of these hacky things it makes you do. Here's a cleaner solution
```ts
const cookies = document.querySelector("#cookies").innerText.split("\n")
presenceData.state = cookies.pop()
presenceData.details = cookies.join(" ")
```
This should work for when the cookie number is too large to fit in 1 line too

websites/D/DashNet/dist/metadata.json
```diff
@@ -4,6 +4,12 @@
 		"name": "Bas950",
 		"id": "241278257335500811"
 	},
+	"contributors": [
+		{
+			"name": "DananaBanana",
+			"id": "423478609529929728"
+		}
+	],
```
ImRodry, line 12: also this PR isn't enough to add yourself as a contributor

PR #6339, opened 2022-06-18
feat(tools): add util file and compileChanged

ImRodry:
ci: send event name to compile## Description 

This PR moves utility functions from all tools to a util.ts folder to avoid code duplication and removes the old massCompile script in favor of compileChanged, which was also added to the CI check to ensure presences compile before being merged.

This PR also removes all instances where https://github.com/PreMiD/changed-files was used, so that repo can be deleted after this PR is merged

# ⚠️ **PLEASE, PLEASE TEST THESE**

## Acknowledgements
- [X] I linted the code by running `yarn format`
- [X] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)


ImRodry:
Proof that CI works as expected: https://github.com/PreMiD/Presences/runs/6951588250?check_suite_focus=true

Bas950:
> Proof that CI works as expected: https://github.com/PreMiD/Presences/runs/6951588250?check_suite_focus=true

image
But it crashed?

.github/workflows/compileLint.yml
```diff
@@ -1,18 +1,24 @@
-name: Lint
+name: Compile and lint
```
EncryptedDev, line 1: This should probably be called Build instead. `build` lints and compiles

EncryptedDev (changes requested):

ImRodry:
> > Proof that CI works as expected: https://github.com/PreMiD/Presences/runs/6951588250?check_suite_focus=true
> 
> image But it crashed?

That’s exactly what I’m trying to prove

.github/workflows/compileLint.yml
ImRodry: It isn’t building anything because it produces no output, it simply checks if stuff works

EncryptedDev: We should think of a better name than this though

.github/workflows/compileLint.yml
```diff
@@ -1,18 +1,24 @@
...
           node-version: 16
           cache: "yarn"
       - name: Install dependencies
         run: yarn
-      - name: Lint
+      - name: Compile changed files
+        run: |
+          git config --global core.quotePath false
+          yarn compile ${{ github.event_name }} $GITHUB_BASE_REF
+      - name: Check Lint
```
EncryptedDev, line 23: Original sounds weird as you're liniting the code
```suggestion
      - name: Linter
```

EncryptedDev (changes requested):

EncryptedDev:
@Bas950 will need to check if this works, I can't use git

.github/workflows/compileLint.yml
ImRodry: You’re checking if the code was linted correctly, you’re not linting it

EncryptedDev: Isn't a linter a lint checker anyway?

Bas950 (approved):

EncryptedDev (approved):

PR #6535, opened 2022-07-29
Update multiple presences (29/07/2022)

Hans5958:
Hello, everyone.

This PR is an accompaniment of the PR #6534 which updates these presences with the same codebase. For more information, please see #6534. Relevant changelogs will be listed here.

This PR includes updates for the following presences:

- Fandom 1.3.0
- Miraheze 1.1.0

## Acknowledgements
- [x] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
- [x] I linted the code by running `yarn format`
- [x] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

## Changelog

- Update base script
  - Update metadata schema to 1.7
  - Change position to save `let`/`const` keyword (Bas950)
  - Adjust to support hiding certain presence data fields

- Adjust codebase from Wikimedia
- Fandom: Update store page logo and thumbnail
- Fandom: Deprecate Fandom/Gamepedia migration code to support all Fandom wikis
- Fandom: Add support for login pages on auth.fandom.com

### Side-by-side

This section only demonstrates Fandom, but it should work on all of the presences, because of the same codebase.

| Image | Link visited |
| ----- | ------------ |
|  | https://fridaynightfunking.fandom.com/wiki/Funkipedia_Mods_Wiki (with button and timestamp) |
|  | https://fridaynightfunking.fandom.com/wiki/Funkipedia_Mods_Wiki (with button only) |
|  | https://fridaynightfunking.fandom.com/wiki/Funkipedia_Mods_Wiki (with timestamps only) |
|  | https://fridaynightfunking.fandom.com/wiki/Funkipedia_Mods_Wiki (without button and timestamp) |

### Variations

| Image | Link visited |
| ----- | ------------ |
|  | https://fridaynightfunking.fandom.com/wiki/Funkipedia_Mods_Wiki |
|  | https://fridaynightfunking.fandom.com/wiki/Plants_vs._Rappers |
|  | https://www.fandom.com/ |
|  | https://terribletvshows.miraheze.org/wiki/Terrible_Shows_%26_Episodes_Wiki |
|  | https://terribletvshows.miraheze.org/wiki/Sonic_Underground |
|  | https://www.miraheze.org/ |

PR #6549, opened 2022-07-29
feat(Fandom): update presence (1.3.0)

Hans5958:
> The title of this PR does not follow the repo's commit conventions

And you can't just do a simple rebase? Come on, man. You can't be serious.

For the full explanation of this PR, please see #6535.

### DIFF
--- a/websites/F/Fandom/dist/metadata.json
+++ b/websites/F/Fandom/dist/metadata.json
@@ -12,9 +12,9 @@
 	},
 	"url": "fandom.com",
 	"regExp": "[a-z0-9-]+[.]fandom[.]com[/]",
-	"version": "1.2.16",
-	"logo": "https://i.imgur.com/sko4P3c.png",
-	"thumbnail": "https://i.imgur.com/VrVSD2b.png",
+	"version": "1.3.0",
+	"logo": "https://i.imgur.com/85TFVjV.png",
+	"thumbnail": "https://i.imgur.com/u7dHJxM.png",
 	"color": "#002A32",
 	"category": "other",
 	"tags": [
@@ -28,5 +28,19 @@
 		"tv",
 		"movies",
 		"anime"
+	],
+	"settings": [
+		{
+			"id": "buttons",
+			"title": "Show Buttons",
+			"icon": "fas fa-compress-arrows-alt",
+			"value": true
+		},
+		{
+			"id": "time",
+			"title": "Show Timestamps",
+			"icon": "fad fa-stopwatch",
+			"value": true
+		}
 	]
 }
--- a/websites/F/Fandom/presence.ts
+++ b/websites/F/Fandom/presence.ts
@@ -1,401 +1,377 @@
-if (
-	document.location.pathname.includes("/wiki/")
-		? document.querySelector(".skin-oasis") ||
-		  ((document.querySelector(".skin-fandomdesktop") ||
-				document.querySelector(".skin-fandommobile")) &&
-				!document.querySelector(".is-gamepedia"))
-		: true
-) {
-	((): void => {
-		const presence = new Presence({
-				clientId: "644400074008297512",
-			}),
-			browsingTimestamp = Math.floor(Date.now() / 1000);
-		let currentURL = new URL(document.location.href),
-			currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/"),
-			presenceData: PresenceData = {
-				details: "Viewing an unsupported page",
-				largeImageKey: "lg",
-				startTimestamp: browsingTimestamp,
-			};
-		const updateCallback = {
-				_function: null as () => void,
-				get function(): () => void {
-					return this._function;
-				},
-				set function(parameter) {
-					this._function = parameter;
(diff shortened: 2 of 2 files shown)


websites/F/Fandom/dist/metadata.json
```diff
@@ -28,5 +28,19 @@
 		"tv",
 		"movies",
 		"anime"
+	],
+	"settings": [
+		{
+			"id": "buttons",
+			"title": "Show buttons",
```
EncryptedDev, line 35:
```suggestion
			"title": "Show Buttons",
```

websites/F/Fandom/dist/metadata.json
```diff
@@ -28,5 +28,19 @@
...
+			"icon": "fas fa-compress-arrows-alt",
+			"value": true
+		},
+		{
+			"id": "time",
+			"title": "Show timestamps",
```
EncryptedDev, line 41:
```suggestion
			"title": "Show Timestamps",
```

websites/F/Fandom/presence.ts
```diff
@@ -1,401 +1,407 @@
...
+		defaultData: PresenceData = {
+			details: "Viewing an unsupported page",
+			largeImageKey: "lg",
+			startTimestamp: browsingStamp,
+		}
+	): void => {
+		currentURL = new URL(document.location.href);
+		currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/");
+		presenceData = { ...defaultData };
+	},
```
EncryptedDev, line 34: Alright, this goes for all the PRs. This seems like an overly sophisticated method of completing a fairly simple task which is fine normally, but as the Presence repo is open-source this needs to be widely understandable and I haven't got a clue on why you have added this level of confusion to a very simple function. 
Please refactor this and make it understandable for everyone.

EncryptedDev (changes requested):

websites/F/Fandom/dist/metadata.json
Hans5958: I saw the PreMiD presence use sentence case and it is a mix on the YouTube presence. 

Are you sure about this? Please confirm again.

websites/F/Fandom/presence.ts
Hans5958: I know this is subjective, but I think confusion is an overstatement 

The resetData function is to reset the presence data to the state after it is prepared. 

After the presence data has been prepared (on the `prepare()` function), the usual way I did is to send the data directly, skipping the preparation part (and an optimal way than conventional ways). However, there are some cases when after the preparation, the page doesn't refresh but changes in some way, you need to update the presence again. For Fandom, this is when a user wants to edit the page with VisualEditor. For that, the `updateCallback()` is made to store the functions on what to do after the preparation. This also true if you want to hide the buttons and the timestamps using the settings.

The way it works is that the function modifies the **prepared** presence data, instead of start over with the blank page, also skipping the heavier parts. With that, on the `presence.on("UpdateData", () => {})` callback, the presence data would be reset to the **prepared** state before the `updateCallback()` can modify it, along with the buttons and timestamps can be hidden based on the settings.

Now it's been a while since I created this code base (and also debated), and with the use case of the MediaWiki sites, I have kind of simplified it. I have simplified this part of code enough, and I hope it is clear enough now.

I'm unsure if this is also your point, but I putting this here: However, it is my best interest to not compromise the quality of the code/logic in favor of readability/understandability. I do recognize a better way to do this: documentation in form of code comments. I would do this within your comment about this.

[Related write-up of my code base, but it's from 2020 so a lot has changed since then.](https://hans5958.github.io/blog/2020/06/29/explanation-for-my-presence-ts-base-on-premid)

EncryptedDev: The confusion is very valid, other's have also questioned it too. I kind of see why you are doing this, but from a maintainability perspective this presence is poor because it isn't really inline with other presences which makes it incredibly hard to maintain

websites/F/Fandom/dist/metadata.json
Hans5958: Edited. By the way, would you like "Show Buttons" or "Show Button", since there is only one on every page. This is also for "Show Timestamps"

ImRodry: show button wouldn't be correct english

EncryptedDev:
Please stop force pushing. Really not helpful for reviewing...

websites/F/Fandom/presence.ts
```diff
@@ -1,401 +1,377 @@
...
+ */
+const resetData = (defaultData: PresenceData): void => {
+		currentURL = new URL(document.location.href);
+		currentPath = currentURL.pathname.replace(/^\/|\/$/g, "").split("/");
+		presenceData = { ...defaultData };
+	},
+	/**
+	 * Search for URL parameters.
+	 * @param urlParam The parameter that you want to know about the value.
+	 */
```
EncryptedDev, line 31: Isn't needed

EncryptedDev (changes requested):

PR #6565, opened 2022-07-31
feat(Tanks.gg): add presence

unicornbetrayal:
## Description 

Added a presence for [tanks.gg](https://tanks.gg/)
Resolves https://github.com/PreMiD/Presences/issues/5314

## Acknowledgements

## Screenshots
<details>

image
image
image
image
image

</details>


EncryptedDev:
Please complete the PR's acknowledgements

websites/T/Tanks.gg/dist/metadata.json
```diff
@@ -0,0 +1,24 @@
...
+      "name": "unicornbetrayal",
+      "id": "377284540420587541"
+    },
+    "service": "Tanks.gg",
+    "description": {
+      "en": "World of Tanks tank info and models."
+    },
+    "url": "tanks.gg",
+    "version": "1.0.0",
+    "logo": "https://i.imgur.com/PjDWjpk.png",
```
EncryptedDev, line 13: Logo should be at least 512 x 512

unicornbetrayal:
> Please complete the PR's acknowledgements

Sorry forgot to click them lol

EncryptedDev (changes requested):
Metadata should also be formatted with tabs (run `yarn format`)

websites/T/Tanks.gg/presence.ts
```diff
@@ -0,0 +1,105 @@
...
+			startTimestamp: browsingTimestamp,
+		},
+		{ pathname, href } = window.location;
+
+	if (pathname === "/") presenceData.details = "Browsing the home page";
+	else if (document.location.pathname.startsWith("/tank/")) {
+		presenceData.details = "Viewing a tank";
+		presenceData.state = `${document
+			.querySelector("head > title")
+			.textContent.replace(" - World of Tanks - tanks.gg", "")}`;
```
DarkVillager, line 18: again why???

unicornbetrayal: I tried to use the selector that included the text I wanted to display, but this error would always come back if you attempted to view a different country, which I couldn't fix unfortunately 
image

RisingSunLight42: Idk what selector you used, but with this
`document.querySelector(".header-tank > h1").childNodes[0].textContent` you can have the name of the tank, and this part of the page don't change with the country I think.
Tested on https://tanks.gg/tank/m-v-y and this work.
Btw, just suggestion, but if this selector works, then you can add if the tank is versatile, etc, and much more, this could be interesting to add more features to your presences.

unicornbetrayal: Oops my bad, I was looking at the tech tree page and code. I'll implement this selector for the tanks, thanks

unicornbetrayal (commented):
Fixed issues and selectors

websites/T/Tanks.gg/presence.ts
unicornbetrayal: @DarkVillager @RisingSunLight42 Thanks to both of you for bringing this to my attention, by the way, it helped me fix an issue that I couldn't figure out before

websites/T/Tanks.gg/presence.ts
```diff
@@ -0,0 +1,117 @@
...
+			{
+				label: "View Page",
+				url: href,
+			},
+		];
+	} else if (document.location.pathname.startsWith("/techtree/")) {
+		presenceData.details = "Viewing the tech tree";
+		presenceData.state = `${
+			document.querySelectorAll(".current")[0].innerHTML
+		}`;
```
DarkVillager, line 42: If you're getting the first one from a querySelectorAll then use a normal querySelector cuz that always gets the first one
```suggestion
		presenceData.state = `${
			document.querySelector(".current").innerHTML
		}`;
```

websites/T/Tanks.gg/presence.ts
```diff
@@ -0,0 +1,117 @@
...
+			largeImageKey: "https://i.imgur.com/jjvJi16.png",
+			startTimestamp: browsingTimestamp,
+		},
+		{ pathname, href } = window.location;
+
+	if (pathname === "/") presenceData.details = "Browsing the home page";
+	else if (document.location.pathname.startsWith("/tank/")) {
+		presenceData.details = "Viewing a tank";
+		presenceData.state = `${
+			document.querySelector(".header-tank > h1").childNodes[0].textContent
```
DarkVillager, line 17:
```suggestion
			document.querySelector(".header-tank > h1").firstElementChild.textContent
```

websites/T/Tanks.gg/presence.ts
```diff
@@ -0,0 +1,117 @@
...
+			document.querySelector(
+				"#content > div > div.container-fluid.main-container > div > div.top-container > div.main-col > h2"
+			).childNodes[0].textContent
+		}`;
+		presenceData.buttons = [
+			{
+				label: "View Profile",
+				url: href,
+			},
+		];
```
DarkVillager, line 95: This way it works for all profiles.
```suggestion
		} else if (document.querySelector('[class="player-stats"]')) {
		presenceData.details = `Viewing profile (${
			document.querySelector('[class="dropdown-item active"]').textContent
		})`;
		presenceData.state = `${
			document.querySelector(
				"#content > div > div.container-fluid.main-container > div > div.top-container > div.main-col > h2"
			).firstChild.textContent
		}`;
		presenceData.buttons = [
			{
				label: "View Profile",
				url: href,
			},
		];
```

DarkVillager (changes requested):

websites/T/Tanks.gg/presence.ts
RisingSunLight42: > 

Nooo !!!!
Use `firstChild` instead ! `firstElementChild` will target the first node nested in the current node targeted, without the texte !
So, for example :
```...<h1>Name <small>second name</small> <span>role</span></h1>```
With `document.querySelector(".header-tank > h1").firstElementChild.textContent` the part targeted will be the `small` element, since with `firstChild`, this will be the text "Name".

So use `firstChild`, DarkVillager is right in the fact of using a specific attributes to get it instead of `childNodes[0]`, I was not aware of `firstChild` when I suggested you the change.

Comment (unicornbetrayal) [reply websites/T/Tanks.gg/presence.ts]: Thanks for the help, guys

File: websites/T/Tanks.gg/presence.ts [hunk 6]
```diff
@@ -37,9 +37,7 @@ presence.on("UpdateData", async () => {
 		];
 	} else if (document.location.pathname.startsWith("/techtree/")) {
 		presenceData.details = "Viewing the tech tree";
-		presenceData.state = `${
-			document.querySelector(".current").innerHTML
-		}`;
+		presenceData.state = `${document.querySelector(".current").innerHTML}`;
```
RisingSunLight42, line 40: Since you're not passing anything in the string other than the query that return a string, this is useless.
Also, prefer to use textContent since innerHTML is less clear when you read and also less fast than textContent.
```suggestion
		presenceData.state = document.querySelector(".current").textContent;
```

RisingSunLight42 (changes requested):

RisingSunLight42 (approved):

EncryptedDev (approved):

DarkVillager (approved):

Bas950 (approved):

PR #5568, opened 2022-01-11
refactor(Gmail): general recode and add localization support

EncryptedDev:
Extends #5216

PR #6583, opened 2022-08-02
refactor(Gmail): redoing codebase and add localization

RisingSunLight42:
## Description 

Refactoring the whole presence by redoing the codebase and adding the localization. I'll add a changelog section for better explanation on some choices.

## Changelog
**I've taken inspiration of PR #5568 by @EncryptedDev on some points, but there's some changes**

- I've added an `pages` Record of strings to avoid long if/else statements or switch/case statements.
> - However, I've decided to not do a for loop since it could be less faster than an access to the object, so there's some if/else, but I think 5 of them is fine
- Also, I found out that there's no localization for when you're viewing your inbox, so I've guessed it was the "primary category", but I'm not sure, so I would like to know if there's a change needed here.
- I've added privacy setting to show or not the label currently viewed, I've also found a good querySelector better than the previous one.
- I've added a show timestamp setting
- And, of course, localization !

I hope the code is good enough to get this PR merged, I'll try to be fast as possible in responding you if you have change requests !

## Acknowledgements

## Screenshots
<details>

gmail_proof_1

gmail_proof_5

gmail_proof_4

gmail_proof_3

gmail_proof_2

</details>

### DIFF
--- a/websites/G/Gmail/dist/metadata.json
+++ b/websites/G/Gmail/dist/metadata.json
@@ -1,13 +1,13 @@
 {
 	"$schema": "https://schemas.premid.app/metadata/1.7",
 	"author": {
-		"name": "Striker",
-		"id": "215509157837537280"
+		"name": "RisingSunLight",
+		"id": "240521747852558347"
 	},
 	"contributors": [
 		{
-			"name": "RisingSunLight",
-			"id": "240521747852558347"
+			"name": "Striker",
+			"id": "215509157837537280"
 		},
 		{
 			"name": "Tellen",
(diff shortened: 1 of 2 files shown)


DarkVillager:
@RisingSunLight42 refacto => refactor?

RisingSunLight42:
> @RisingSunLight42 refacto => refactor?

Yup! Woopsy doopsy. Thanks!

EncryptedDev (approved):

RisingSunLight42:
Note : I've edited the behavior of the `compose` part since the compose=new is not consistent in the URL, so I've just fixed it by searching `compose=` in the URL with `includes` instead of `endswith`

Bas950 (approved):

PR #6623, opened 2022-08-09
feat(yandex.ru): add presence

Nolo001-Aha:
## Description 

Add a presence for Yandex (a search engine popular in Russia and CIS countries). Presence shows the user's current search query/activity across the following services:

- Yandex Search
- Yandex Images
- Yandex Videos
- Yandex News
- Yandex Maps

## Acknowledgements

## Screenshots
<details>

Screenshot 1

Screenshot 2

Screenshot 3

</details>


DarkVillager:
@Nolo001-Aha use variables for frequently accessed code

RisingSunLight42:
Also, maybe use a switch statement instead of a if statement (in combination of the previous comment by Dark)

Nolo001-Aha:
> Also, maybe use a switch statement instead of a if statement (in combination of the previous comment by Dark)

I think using if/else if is better in this situation because I use the **startsWith** method, and it returns a bool. I'd have to make sure I trim the query string so that I can have explicit **case**s, and wouldn't that be less effective?

DarkVillager:
> > Also, maybe use a switch statement instead of a if statement (in combination of the previous comment by Dark)
> 
> I think using if/else is better in this situation because I use the **startsWith** method, and it returns a bool. I'd have to make sure I trim the query string so that I can have explicit **case**s, and wouldn't that be less effective?

or just use a switch statement that check `document.location.pathname.split("/")[1]`
image

DarkVillager (approved):

websites/Y/Yandex/presence.ts
```diff
@@ -2,25 +2,33 @@ const presence = new Presence({
...
+		case "image": {
+			presenceData.details = `Looking for Images of ${textQuery}`;
+			break;
+		}
+		default:
+			if (currentPath[1] === "video" && currentPath[2] === "search")
+				presenceData.details = `Looking for Videos of ${textQuery}`;
+			else if (currentPath[1] === "news")
+				presenceData.details = `Looking for News on ${textQuery}`;
+			else if (currentPath[1] === "maps") presenceData.details = "Using Maps";
```
RisingSunLight42, line 31: Why don't you put them in case ?

Nolo001-Aha:
Yeah sorry, sometimes using "Fix all - Convert..." isn't a good idea. Fixed and checked, should be good to go now (I hope). Sorry for the inconveniences, my first upload here.

EncryptedDev:
Hello there 👋🏼

Thank you for your contribution to the store. But at this current time, we will not proceed with this pull request as I believe it is not at a high enough quality to add to the store. 

There isn't anything wrong with the pull request, but there is just not much to the presence and should add more pages if possible - if this isn't possible because the service doesn't support more, that is fine and let me know so I can reopen the pull request. 

If you need anything, reach out to me on Discord - Encrypted#0001

PR #6653, opened 2022-08-13
feat: add automatically generating codeowners

EncryptedDev:
Co-Authored-By: Rodry <38259440+ImRodry@users.noreply.github.com>
## Description 

- Automatically generate `CODEOWNERS` file, by new `1.8` Schema. (PreMiD/Schemas#101)
- Rename `#` to `*` folders due to conflicts with `CODEOWNERS` file
- Rename workflows to use dashes
- Bump schemas to `1.8`


Bas950:
image
And shouldn't you still be added as a reviewer always?

ImRodry:
The codeowners file won’t actually work as codeowners, it’s simply so you can then make a script that assigns those people. And no you should only be assigned if there are no contributors

PR #6819, opened 2022-09-25
chore: update dependencies

Timeraa:
This Pull Request updates all dependencies to their latest possible version but also pins them to that version to prevent other package managers like `pnpm` to use the latest version as no lock file exists.

This PR also updates `pmd` to version 1.0.3 which fixes the directory where the metadata.json is placed.


EncryptedDev:
Why would we pin these versions? That would prevent future deps being updated by Dependabot, no?

Timeraa:
No, dependabot will still open prs but package managers won't just use the latest version

Timeraa:
> Why would we pin these versions? That would prevent future deps being updated by Dependabot, no?

No, it will only prevent package managers from using the latest available. Dependabot (soon renovate) will still open Pull Requests.

Timeraa:
For example `pnpm` or `npm` use the latest mongodb version which then caused the mongodb type error.

Bas950 (approved):

EncryptedDev (approved):

PR #6870, opened 2022-10-12
chore: remove dependabot, add renovate

Timeraa:
## Description 
This Pull Request removes dependabot in favor of Renovate. This is because Renovate offers a lot more features and allows us to maintain a single config which we can extend everywhere.

Automerge has been disabled for now, as we do not have automatic tests in this repository that ensure that the updated dependency does not break things.


Bas950 (approved):

EncryptedDev (approved):

EncryptedDev:
For the love of God, don't enable automerge ever - ty

PR #6890, opened 2022-10-15
feat: add sourcemaps

Timeraa:
This PR enables sourceMap generation. This feature was never enabled with unknown reason? Enabling it will help Presence developers and or users with Presence errors diagnose these and fix them a lot easier or well possible in the first place.

Yes, it does make the compiled code of Presences bigger but ultimatively allows devs to debug when users are experiencing issues.


Bas950 (approved):

EncryptedDev (approved):

EncryptedDev:
nooo!!!

PR #6923, opened 2022-10-21
feat(Jackbox): add details for individual games

theusaf:
## Description 

* Adds details for individual Jackbox games
* Adds games up to Party Pack 9
* Adds icons for each game

## Acknowledgements

## Screenshots
<details>

</details>

## Possible issue:
Sometimes, the resulting compiled code cannot be executed using `chrome.tabs.executeScript` possibly due to how large it is.
However, this seems to go away if the compiled code ends with a newline character (tested on firefox and chrome)
* This behavior seems to happen randomly but is consistent when it starts happening.


Bas950:
@theusaf you need to remove the .js from your imports

theusaf:
Alright, it compiles successfully. However, the issue mentioned in the description of this PR still happens.
Unless a newline character is appended to the end of the generated presence.js file, the code doesn't run properly for some reason.

Bas950:
@Timeraa could you look into that

theusaf:
> Alright, it compiles successfully. However, the issue mentioned in the description of this PR still happens. Unless a newline character is appended to the end of the generated presence.js file, the code doesn't run properly for some reason.

I figured out the cause. Basically, the source map comment, [when combining the presence and logger code](https://github.com/PreMiD/Extension/blob/5e9894de2c5e82d9425123ab353859158413e65d/src/util/functions/injectPresence.ts#L47-L48) comments out the first line of the logger injection. This breaks reading logs.

Bas950:
> > Alright, it compiles successfully. However, the issue mentioned in the description of this PR still happens. Unless a newline character is appended to the end of the generated presence.js file, the code doesn't run properly for some reason.
> 
> I figured out the cause. Basically, the source map comment, [when the extension combines the presence and logger code](https://github.com/PreMiD/Extension/blob/5e9894de2c5e82d9425123ab353859158413e65d/src/util/functions/injectPresence.ts#L47-L48), comments out the first line of the logger. This breaks reading logs.

Amazingly found! @Timeraa will look into it further in the coming days. This is now internally tracked under code: PMD-90

Timeraa:
I suppose the fox would be adding an empty line in the compiler hmm

EncryptedDev:
Hi @theusaf, sorry for the delay here. Trying to work out the best way of reviewing this as it is quite hard to do so

websites/J/Jackbox/games/Party Pack 3/awshirt.ts
```diff
@@ -0,0 +1,70 @@
...
+			presenceData.state = "Answering a question";
+			break;
+		}
+		case "state-done-answering": {
+			presenceData.state = "Waiting for other players to finish answering";
+			break;
+		}
+	}
+	return presenceData;
+}
```
Bas950, line 70: I think the return {state:...} would look better here too, would say a lot of space by  removing all those breaks.

websites/J/Jackbox/games/Party Pack 3/fakinit.ts
```diff
@@ -0,0 +1,28 @@
...
+	else if (classList.contains("state-notchoosing"))
+		presenceData.state = "Waiting for a category to be chosen";
+	else if (classList.contains("state-round"))
+		presenceData.state = currentGamePage.querySelector("p").textContent;
+	else if (classList.contains("state-vote"))
+		presenceData.state = "Voting on a player";
+	else if (classList.contains("state-vote-locked"))
+		presenceData.state = "Viewing vote results";
+	return presenceData;
+}
```
Bas950, line 28: My personal preference but I think this could look cleaner in a switch (true) case basis, but @EncryptedDev on that.

And then in the switch use the return stuff like before.

Bas950 (changes requested):
Love the use of your own @types.

Those 2 comments have the same structure on more files, so please go through others too and make the syntax you use the same for consistency.

websites/J/Jackbox/games/Party Pack 3/awshirt.ts
theusaf: I've applied this style to all files without a small image in the presence data.

EncryptedDev (dismissed):
lgtm

Bas950:
> lgtm

Please take a look at the conversations where I tagged ya in..

Bas950 (approved):

EncryptedDev (approved):

PR #6945, opened 2022-10-28
chore: set moduleResolution to nodenext

ImRodry:
Ref: https://www.typescriptlang.org/tsconfig#moduleResolution

## Description 

turns out module esnext needs this I think?


EncryptedDev:
ref ref

PR #6946, opened 2022-10-29
fix(presenceUpdater): use api types in compiler options

theusaf:
## Description 

* Should fix compilation of presences.
* See https://github.com/microsoft/TypeScript/issues/23564#issuecomment-383092947

## Acknowledgements

### DIFF
--- a/util/tools/auto/presenceUpdater.ts
+++ b/util/tools/auto/presenceUpdater.ts
@@ -17,8 +17,12 @@ import { valid } from "semver";
 import { minify as terser } from "terser";
 import typescript from "typescript";
 import { type CompilerOptions } from "typescript";
-const { createProgram, flattenDiagnosticMessageText, getPreEmitDiagnostics } =
-	typescript;
+const {
+	createProgram,
+	flattenDiagnosticMessageText,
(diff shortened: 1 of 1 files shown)


Slowlife01:
Needs to do this for `target` too. I tested it.
And why not just hard code it?
image

theusaf:
> Needs to do this for `target` too. I tested it. And why not just hard code it? image

Strangely, I didn't have that problem with `target`, but I'll add that just in case. The reason I'm not hardcoding it is so that only the `tsconfig.json` file needs to be changed to change the `target`/`moduleResolution`

Slowlife01:
> > Needs to do this for `target` too. I tested it. And why not just hard code it? image
> 
> Strangely, I didn't have that problem with `target`, but I'll add that just in case.

This is just to fix to warnings with stuff like `String.replceAll()`

> The reason I'm not hardcoding it is so that only the `tsconfig.json` file needs to be changed to change the `target`/`moduleResolution`
I figured that would be your reason. Whatever, I guess

theusaf:
Closing as CI was updated, which appears to fix this issue.

PR #6988, opened 2022-11-10
fix(presenceUpdater): correctly detect changed folders

theusaf:
## Description 

* Corrects the order of sources in getting diffs
  * Allows new presences to be detected and uploaded
  * Allows deleted presences to be detected and deleted
* Adds a check for changed presences before writing
  * Avoids preventing a presence from being deleted

## Acknowledgements
- [x] I linted the code by running `yarn format`
- [x] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

PR #6989, opened 2022-11-11
chore: bump presences

EncryptedDev:
bump presences to fix CI for presences for the last 22 days; since my last bulk PR

should be the final time this is needed due to #6988


EncryptedDev:
CI is going to fail😁

PR #7030, opened 2022-11-20
chore: fix presenceValidator

Slowlife01:
image
image


Slowlife01:
If path array is longer than 2, it wouldn't be able to get there. Will add support for that *just in case*

Bas950:
Could you add a before - after pic

Slowlife01:
> Could you add a before - after pic
> 
> Don't quite get what has changed.

image
image

Bas950 (approved):

EncryptedDev (approved):

PR #7050, opened 2022-12-01
refactor: monthly SE

SeMiD:
*beep boop* I've prettified and refactored some Presences and sorted all metadata files...


Slowlife01:
Probably should add uncommitted to the argument since it uses getDiff()

PR #7068, opened 2022-12-11
Update YandexMusic

cheburashkalev:
largeImageKey patch for show image track image in discord

Signed-off-by: Cheburashka <47310777+cheburashkalev@users.noreply.github.com>

Screenshot 2022-12-12 025332

### DIFF
--- a/websites/Y/Yandex Music/presence.ts
+++ b/websites/Y/Yandex Music/presence.ts
@@ -37,7 +37,7 @@ setInterval(async () => {
 	}
 
 	presenceData = {
-		largeImageKey: "og-image",
+		largeImageKey: (<HTMLImageElement>document.querySelector("body > div.page-root.page-root_no-player.deco-pane-back.theme.theme_dark.black > div.bar > div.bar__content > div.player-controls.deco-player-controls > div.player-controls__track-container > div > div > div.track.track_type_player > div.entity-cover.deco-entity-image-placeholder.track-cover.entity-cover_size_Large > a > img")).src.replace("50x50","100x100"),
 		smallImageKey: playing ? "play" : "pause",
 		smallImageText: playing ? (await strings).playing : (await strings).pause,
 		details: (document.querySelectorAll(".track__title")[0] as HTMLElement)


websites/Y/Yandex Music/presence.ts
```diff
@@ -37,7 +37,7 @@ setInterval(async () => {
 	}
 
 	presenceData = {
-		largeImageKey: "og-image",
+		largeImageKey: (<HTMLImageElement>document.querySelector("body > div.page-root.page-root_no-player.deco-pane-back.theme.theme_dark.black > div.bar > div.bar__content > div.player-controls.deco-player-controls > div.player-controls__track-container > div > div > div.track.track_type_player > div.entity-cover.deco-entity-image-placeholder.track-cover.entity-cover_size_Large > a > img")).src.replace("50x50","100x100"),
```
theusaf, line 40: Prefer an Imgur URL instead of this very long query selector, which can change frequently.

websites/Y/Yandex Music/presence.ts
```diff
@@ -37,7 +37,7 @@ setInterval(async () => {
...
 		smallImageKey: playing ? "play" : "pause",
```
theusaf, line 41: Also, after replacing the `largeImageKey` with an Imgur URL, also replace all the `smallImageKey`s with Imgur URLs as well, following the `enum` pattern found in Presences like `Crunchyroll`

theusaf (changes requested):
Resolve CI.

websites/Y/Yandex Music/presence.ts
Slowlife01: Pretty sure they wants to use cover art..
You can use `navigator.mediaSession.metadata.artwork.at(-1).src` instead.

Slowlife01: Pretty sure they wants to use cover art..
You can use `navigator.mediaSession.metadata.artwork.at(-1).src` instead.

PR #7213, opened 2023-03-05
feat(ticketer): add presence

mezotv:
## Description 
A rich presence addon for ticketer. A ticket bot with a web dashboard for discord.

## Acknowledgements
 * [x] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
 * [x] I linted the code by running `yarn format`
 * [x] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

 ## Screenshots
 Proof showing the creation/modification is working as expected

image
image

### DIFF
--- a/websites/T/Ticketer/metadata.json
+++ b/websites/T/Ticketer/metadata.json
@@ -11,7 +11,7 @@
 	},
 	"url": "ticketerbot.com",
 	"version": "1.0.0",
-	"logo": "https://i.imgur.com/2dILrRV.png",
+	"logo": "https://i.imgur.com/XDt1aUu.png",
 	"thumbnail": "https://i.imgur.com/HoFreB1.png",
 	"color": "#5865f2",
 	"category": "other",
--- a/websites/T/Ticketer/metadata.json
+++ b/websites/T/Ticketer/metadata.json
@@ -2,7 +2,7 @@
 	"$schema": "https://schemas.premid.app/metadata/1.7",
 	"author": {
 		"id": "347077478726238228",
-		"name": "Dominik#5555"
+		"name": "Dominik"
 	},
 	"service": "Ticketer",
 	"description": {
--- a/websites/T/Ticketer/metadata.json
+++ b/websites/T/Ticketer/metadata.json
@@ -1,23 +1,23 @@
 {
-  "$schema": "https://schemas.premid.app/metadata/1.7",
-  "service": "Ticketer",
-  "author": {
-    "id": "347077478726238228",
-    "name": "Dominik#5555"
-  },
-  "description": {
-    "en": "Simplifies ticketing for customer support, sales, and suggestions",
-    "de": "Vereinfacht das Ticketing für Kundensupport, Verkauf und Vorschläge"
-  },
-  "url": "ticketerbot.com",
-  "version": "1.0.0",
-  "logo": "https://i.imgur.com/2dILrRV.png",
-  "thumbnail": "https://i.imgur.com/HoFreB1.png",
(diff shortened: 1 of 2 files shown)


mezotv:
Will add that null check tomorrow! Can't think straight right now.

DarkVillager:
The website is down?
image

mezotv:
> The website is down? image

Probably only a temporary issue. Should be up soon!

DarkVillager (changes requested):

websites/T/Ticketer/presence.ts
```diff
@@ -0,0 +1,63 @@
...
+			if (document.querySelector(".server-info > h3")) {
+				title = document.querySelector(".server-info > h3");
+				logo = document.querySelector(".server-info > img");
+				dashboardInfo = document.querySelector(".informations-nav > h2");
+
+				presenceData.details = "Server Settings - Editing:";
+				presenceData.state = title.textContent;
+				presenceData.smallImageKey = logo.getAttribute("src");
+			}
+			if (!dashboardInfo?.textContent) return;
```
DarkVillager, line 40: You could just give the text "Browsing the dashboard"?
```suggestion
			if (!dashboardInfo?.textContent) return;
```

DarkVillager (changes requested):

websites/T/Ticketer/presence.ts
mezotv: Yea that sounds much better! Will change that real quick.

websites/T/Ticketer/presence.ts
```diff
@@ -0,0 +1,63 @@
...
+		case "/changelog": {
+			presenceData.details = "Viewing the changelog-page";
+			break;
+		}
+		case "/contribute": {
+			presenceData.details = "Viewing the contribution-page";
+			break;
+		}
+		case "/dashboard": {
+			presenceData.details = "Browsing the dashboard";
```
RisingSunLight42, line 27: Would be better if those case are in a Record like here : https://github.com/PreMiD/Presences/blob/main/websites/V/Voiranime/presence.ts

mezotv: I kine like the way it currently is easy to read. That just seems like an extra step.

mezotv: I tried adding Records like you but they would require way to much code change and I think its easier if I stick to my simple solution!

mezotv:
Not sure why the first check failed its the same code but I removed the - from the strings.

mezotv:
> rate limit

So what do I do now?

Kuriel23:
> So what do I do now?

Nothing, you don't have to worry about it when you give rate limit.

websites/T/Ticketer/metadata.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://schemas.premid.app/metadata/1.7",
+  "service": "Ticketer",
+  "author": {
+    "id": "347077478726238228",
+    "name": "Dominik#5555"
```
Kuriel23: Remove the discriminator

mezotv: Done!

DarkVillager (approved):

RisingSunLight42 (approved):

Bas950 (approved):

Timeraa (approved):

PR #7288, opened 2023-04-13
feat(dアニメストア): update presence

neco222:
I tried changing the code.
I don't know if it works, but could you please check?

The revised code changes the method of retrieving the video element of the player and now retrieves #dplayer_video_box video instead of #video. It has also been modified to set the start timestamp only if the video is playing. Finally, presence.setActivity() has been added to clear the presence if the player's video element is not present.

### DIFF
--- a/websites/D/dアニメストア/presence.ts
+++ b/websites/D/dアニメストア/presence.ts
@@ -1,36 +1,29 @@
-{
-	const presence = new Presence({
-			clientId: "611012705306017792",
-		}),
-		strings = presence.getStrings({
-			play: "general.playing",
-			pause: "general.paused",
-		});
+const presence = new Presence({
+  clientId: "611012705306017792"
+});
 
-	presence.on("UpdateData", async () => {
-		if (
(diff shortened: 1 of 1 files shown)


DarkVillager:
This code doesn't work correctly. If you close this PR, I will change the code to work.

websites/D/dアニメストア/presence.ts
```diff
@@ -1,36 +1,34 @@
...
 
-			presence.setActivity(presenceData);
-		}
-	});
-}
+    // ビデオが存在している場合
+    if (video) {
+        const presenceData: PresenceData = {
+            details: "Watching a video on dアニメストア",
+            state: video.currentTime > 0 && !video.paused ? `Playing` : `Paused`,
```
Bas950, line 16: This should probably show what they are watching.

Bas950 (changes requested):
You have to test the presence yourself and upload screenshots in your PR as proof, to learn more about loading your presence and presence creation as a whole please see: https://docs.premid.app/en/dev/presence#loading-your-presence

PR #7320, opened 2023-04-29
chore(viu): fix eslint warning

DarkVillager:
## Description 

Fix eslint warning
## Acknowledgements
- [x] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
- [x] I linted the code by running `yarn format`
- [x] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>
<summary> Proof showing the creation/modification is working as expected </summary>

n/a

</details>

PR #7326, opened 2023-05-01
refactor: monthly SE

SeMiD:
*beep boop* I've prettified and refactored some Presences and sorted all metadata files...

### DIFF
--- a/websites/V/Viu/presence.ts
+++ b/websites/V/Viu/presence.ts
@@ -90,7 +90,7 @@ async function generateToken() {
 	return response.json().then(x => x.token);
 }
 
-/* eslint-disable camelcase */
+ 
 interface SeriesInfo {
 	is_movie: number;
 	series_id: string;
--- a/websites/A/AnimeVost/metadata.json
+++ b/websites/A/AnimeVost/metadata.json
@@ -17,7 +17,7 @@
 		"www.animevost.org",
 		"v2.vost.pw"
 	],
-	"version": "1.0.1",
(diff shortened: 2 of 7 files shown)


websites/V/Viu/presence.ts
```diff
@@ -90,7 +90,7 @@ async function generateToken() {
 	return response.json().then(x => x.token);
 }
 
-/* eslint-disable camelcase */
```
DarkVillager, line 93: @Bas950 this should not be done.

DarkVillager (changes requested):

websites/V/Viu/presence.ts
Bas950: Hmm we should check why it even removes it

theusaf: image

After removing the comment, eslint does not actually throw any warnings.

theusaf: It looks like eslint ignores camelcase 'issues' in interfaces.

Bas950: so then it can be removed?

theusaf: Yes

DarkVillager: > image
> 
> After removing the comment, eslint does not actually throw any warnings.

I tested it with #7320 it gives deepscan errors

Slowlife01: I added it because of Deepscan

PR #7352, opened 2023-05-13
fix(svtplay): update query selectors

whallin:
## Description 

Since the current presence version of the website doesn't work with the old query selectors, they have now been fixed to work with the latest changes to the SVT Play website.

## Acknowledgements

## Screenshots
<details>

### Watching via "/video/"
Showcasing "watching via /video/"

### Watching live via "/video/"

> COMMENT: SVT Play seems to now use a set duration/timecode when watching live broadcasts via "/video/" instead of showing how much time has passed since the broadcast started. This means that for live material watched through the "/video/" path, the "fetched" time will be a "left" time.

Showcasing "live via /video/"

### Live via "/kanaler/"
Showcasing "live via kanaler"

### Search
Showcasing "search"
</details>

### DIFF
--- a/websites/S/SVT Play/presence.ts
+++ b/websites/S/SVT Play/presence.ts
@@ -70,9 +70,7 @@ presence.on("UpdateData", async () => {
 			title = document.querySelector(
 				'[data-rt="program-info-title"] > a'
 			).textContent;
-			user = document.querySelector(
-				'[data-rt="episode-link"]'
-			).textContent;
+			user = document.querySelector('[data-rt="episode-link"]').textContent;
 			if (video) {
 				if (!video.duration) {
 					time = false;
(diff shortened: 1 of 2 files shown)


websites/S/SVT Play/presence.ts
```diff
@@ -24,13 +24,13 @@ presence.on("UpdateData", async () => {
 			presenceData.details = "Navigerar program kategorier";
 		} else if (document.location.pathname.includes("/kanaler/")) {
 			const video = document.querySelector<HTMLVideoElement>(
-				"#play_main-content > article > div.play_channels-video--show > div > div > div.play_video-player.lp_video.play_channels__active-video > div > video"
+				'[data-rt="video-player-channels"]'
 			);
 			title = document.querySelector(
-				"#play_main-content > article > div.play_channels-video--show > div > div > div.play_channels__active-video-info > h2"
+				"#play_main-content > div > div.sc-b35938a7-1.bTCBgw > div > div > div.sc-2c367a6f-0.dxLhHV > h2"
```
theusaf, line 30: See if there is a more general selector that can be used here for presence stability

whallin: I'm not sure if this would be general enough, but at the very least, it would be better than going after specific classes—that's for sure.

```suggestion
				"#play_main-content > div > div > div > div > div > h2"
```

theusaf (approved):

PR #7378, opened 2023-05-24
Update metadata.json

LieutenantSugar:
HBO Max recently changed its service name to Max. The supported URLs and logo need to be updated accordingly, as the presence in its current state is not detected by PreMiD. If you happen to find a more suitable logo image, feel free to change it. I have not changed anything to the original code. The only thing I modified was the supported URLs and logo image link.

## Acknowledgements
- [x] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
- [ ] I linted the code by running `yarn format`
- [ ] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>
<summary> Proof showing the creation/modification is working as expected </summary>

</details>

PR #7384, opened 2023-05-28
Update HBO Max supported URLs

LieutenantSugar:
I made corrections to #7378 regarding the logo URL. It is from Imgur, so it passes. Otherwise, I have not changed any original code, only the supported URLs and logo URL.

### DIFF
--- a/websites/H/HBO Max/metadata.json
+++ b/websites/H/HBO Max/metadata.json
@@ -11,10 +11,10 @@
 		"vi_VN": "Nền tảng phát trực tuyến đem tất cả của HBO với nhiều phim và chương trình truyền hình hơn nữa, thêm các bộ Max Originals."
 	},
 	"url": [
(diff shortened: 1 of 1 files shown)


DarkVillager:
Some countries don’t have max yet and still use hbo max like mine.

LieutenantSugar:
> Some countries don’t have max yet and still use hbo max like mine. IMG_3477 IMG_3478


Perhaps the new "Max" links can be added to the supported URLs, not replacing the old ones. At least until the new change is implemented for every country.

Bas950 (changes requested):
Just like @DarkVillager said, please keep the _old_ urls for now since some areas aren't switched over _yet_

theusaf:
Closing for inactivity. You may open a new PR after resolving the issues mentioned in this one.

PR #7388, opened 2023-05-29
feat(AssetsManager): automatically upload assets to Recodive's CDN

Bas950:
## Acknowledgements

## Screenshots
<details>

</details>


.github/workflows/AssetsUpdater.yml
```diff
@@ -0,0 +1,37 @@
+name: Assets Updater
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "**.ts"
+      - "**.json"
```
Timeraa, line 8: Shouldn't `*.ts` do the trick already or am I wrong?

.github/workflows/AssetsUpdater.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  update:
+    if: github.repository_owner == 'PreMiD'
```
Timeraa, line 11: Don't think this is needed because Workflows are disabled by default on forks now

.github/workflows/AssetsUpdater.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: Checkout Project
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Use Node.js 16
+        uses: actions/setup-node@v3.5.1
+        with:
+          node-version: 16
+          cache: npm
```
Timeraa, line 23: Maybe we should update to LTS at this point?

tools/auto/assetsUpdater.ts
```diff
@@ -0,0 +1,86 @@
...
+			if (errors.length) actions.setFailed("Failed to delete assets");
+		}
+
+		if (toBeMoved.size) {
+			actions.info(
+				chalk.green(`Moving ${toBeMoved.size} asset(s) for ${folder}...`)
+			);
+			const errors: string[] = [];
+			errors.push(...(await assetsManager.uploadAssets(toBeMoved)));
+			errors.push(...(await assetsManager.deleteAssets([...toBeMoved.keys()])));
```
Timeraa, line 40: Could be merged into one .push

tools/classes/AssetsManager.ts
```diff
@@ -0,0 +1,438 @@
...
+
+	getPresenceFolder() {
+		return glob
+			.sync(`{websites,programs}/**/${this.service}`)[0]
+			.replace(/\\/g, "/");
+	}
+
+	getFileExtension(url: string) {
+		return extname(url);
+	}
```
Timeraa, line 52: Why not use `extname` directly instead of that function?

tools/classes/AssetsManager.ts
```diff
@@ -0,0 +1,438 @@
...
+		logo: string;
+		thumbnail: string;
+		assets: Set<string>;
+	} {
+		const assets = new Set<string>(),
+			{ logo, thumbnail } = this.metadata,
+			files = this.allTsFiles;
+
+		for (const tsfile of files) {
+			const file = readFileSync(tsfile, "utf8");
```
Timeraa, line 85: Maybe don't block the thread? (Doesn't matter in this case but since everything is async might as well)

tools/classes/AssetsManager.ts
```diff
@@ -0,0 +1,438 @@
...
+				case ".jpg":
+				case ".jpeg":
+					mimeType = "image/jpeg";
+					break;
+				case ".gif":
+					mimeType = "image/gif";
+					break;
+				case ".webp":
+					mimeType = "image/webp";
+					break;
```
Timeraa, line 285: `mime-types` module would be a better fit there and then just disallowing other ones than "image/xyz" or hardcoding some mimes

tools/classes/AssetsManager.ts
```diff
@@ -0,0 +1,438 @@
...
+								background: { r: 0, g: 0, b: 0, alpha: 0 },
+							})
+							.toFile(newFileLocation);
+						finalFileLocation = newFileLocation;
+					} catch (error) {
+						errors.push(`Error while resizing asset ${url}: ${error.message}`);
+						continue;
+					}
+				}
+			}
```
Timeraa, line 326: Maybe we should do the same for thumbnails (With a different sizing proportion tho? Idk how we would prevent stretching tho)

Timeraa (changes requested):
Requested some changes, they are all not required in my opinion but a nice to have

tools/classes/AssetsManager.ts
Bas950: Can't use async in getters.

.github/workflows/AssetsUpdater.yml
Bas950: Copied from other workflows we have, we should change at a later point

.github/workflows/AssetsUpdater.yml
Bas950: Copied from other workflows we have, we should change at a later point

.github/workflows/AssetsUpdater.yml
Bas950: Copied from other workflows we have, we should change at a later point

tools/classes/AssetsManager.ts
theusaf: Getters can still return Promises, right?

theusaf (approved):

PR #7443, opened 2023-06-25
fix: Obtain streamer name from path rather than HTML Element

Walledgarden:
## Description 
Currently the getElement selector for obtaining the streamers name does not work reliable on my end. Thus, obtaining it from the path is a cleaner and more reliable solution, as this does not depend on any DOM Content to load.

There is no open issue for this behaviour.

## Acknowledgements

### DIFF
--- a/websites/T/Twitch/presence.ts
+++ b/websites/T/Twitch/presence.ts
@@ -420,7 +420,7 @@ presence.on("UpdateData", async () => {
 
 			if (path.includes("/moderator/")) {
 				presenceData.details = strings.modStreamer;
-				presenceData.state = getElement(".stream-info-card p > a");
+				presenceData.state = path.match(/moderator\/([a-z\d][\w]{2,24})$/i)?.[1] ?? 'Not selected';
 
 				if (getElement(".modview-dock-widget p") !== "Offline") {
 					presenceData.smallImageKey = "live";


theusaf:
Can you provide screenshots showing that the change still works/keeps the behavior.

websites/T/Twitch/presence.ts
```diff
@@ -420,7 +420,8 @@ presence.on("UpdateData", async () => {
 
 			if (path.includes("/moderator/")) {
 				presenceData.details = strings.modStreamer;
-				presenceData.state = getElement(".stream-info-card p > a");
+				presenceData.state =
+					path.match(/moderator\/([a-z\d][\w]{2,24})$/i)?.[1] ?? "Not selected";
```
DarkVillager, line 424: There is one more method of getting which streamer it is.
```suggestion
					path.match(/moderator\/([a-z\d][\w]{2,24})$/i)?.[1] ?? document.querySelector('[data-a-target="player-info-title"]')?.textContent ?? "Not selected";
```
image

DarkVillager (changes requested):

Bas950 (changes requested):
Run formatter

Bas950:
> Can you provide screenshots showing that the change still works/keeps the behavior.

This too

PR #7699, opened 2023-10-24
feat(Discord SFX): add features

chimpdev:
## Description 
/users/:id pages have been added to the site. So I create this PR to ensure that the presence handles the new pages correctly.

## Acknowledgements

## Screenshots
<details>

image
</details>


websites/D/Discord SFX/presence.ts
```diff
@@ -71,20 +71,33 @@ presence.on("UpdateData", async () => {
...
 				url: `https://discordsfx.com/users/${pathname.split("/")[2]}`,
-			}
+			},
 		];
+
+		const { searchParams } = new URL(
+			document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
+				?.src || ""
+		);
+		presenceData.largeImageKey = searchParams.get("url");
```
chimpdev, line 100: Also I changed the value of largeImageKey with the avatar of the user.

websites/D/Discord SFX/metadata.json
```diff
@@ -9,9 +9,9 @@
 		"en": "Discord SFX lists some sound files uploaded by other site users that you can download and use for the soundboard of your Discord server, and also allows you to share your own."
 	},
 	"url": "discordsfx.com",
-	"version": "1.0.1",
-	"logo": "https://cdn.rcd.gg/PreMiD/websites/D/Discord%20SFX/assets/logo.png",
-	"thumbnail": "https://cdn.rcd.gg/PreMiD/websites/D/Discord%20SFX/assets/thumbnail.png",
+	"version": "1.1.0",
+	"logo": "https://i.imgur.com/T6CpbXk.png",
+	"thumbnail": "https://i.imgur.com/xaHioYY.png",
```
theusaf, line 14: I think these changes are unneeded.

theusaf (changes requested):

websites/D/Discord SFX/presence.ts
```diff
@@ -70,6 +70,36 @@ presence.on("UpdateData", async () => {
...
+				} Sounds`,
+				url: `https://discordsfx.com/users/${pathname.split("/")[2]}`,
+			},
+			{
+				label: `Received ${
+					document.querySelector<HTMLSpanElement>(
+						'[data-premid-id="Downloads Received"'
+					)?.textContent
+				} Downloads`,
+				url: `https://discordsfx.com/users/${pathname.split("/")[2]}`,
```
theusaf, line 92: Labels should be action items "View Sounds", etc. They should not be used to display stuff you [couldn't fit in other fields](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md#presencets).

websites/D/Discord SFX/presence.ts
theusaf: This could potentially lead to an empty large image key.

theusaf (changes requested):

websites/D/Discord SFX/presence.ts
```diff
@@ -70,6 +70,19 @@ presence.on("UpdateData", async () => {
...
+		presenceData.details = `Viewing ${
+			document.title.split(" Profile")[0].split(" | ")[1]
+		} profile`;
+
+		const { searchParams } = new URL(
+			document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
+				?.src ||
+				"https://cdn.rcd.gg/PreMiD/websites/D/Discord%20SFX/assets/logo.png"
+		);
+		presenceData.largeImageKey = searchParams.get("url");
```
theusaf, line 83: This still has the same issue. I was wrong in my initial request. If the image element is not found, it would've errored out, rather than resulting in an empty largeImageKey. Either add some null checks, or use a `try/catch` block

theusaf (changes requested):

websites/D/Discord SFX/presence.ts
chimpdev: I understand what you're saying. I'll make a few changes. Also keep in mind there won't be a situation where the image element can't be found. There will always be an image element on the page

theusaf: I see. I was assuming that there could be a chance that there is no image, since you were adding a null check on the `.src` property. If the image is always there, then
```suggestion
		const { searchParams } = new URL(
			document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
				.src
		);
		presenceData.largeImageKey = searchParams.get("url");
```
might be more clear.

chimpdev: I did the null check because prettier told me to do it

websites/D/Discord SFX/presence.ts
```diff
@@ -3,10 +3,12 @@ const presence = new Presence({
 });
 
 presence.on("UpdateData", async () => {
-	const presenceData: PresenceData = {
+	const assets = {
+			Logo: "https://cdn.rcd.gg/PreMiD/websites/D/Discord%20SFX/assets/logo.png",
+		},
```
theusaf, line 8: For consistency with other presences, let's put this in an enum like so:
https://github.com/PreMiD/Presences/blob/1f5a4117f92c6549a75e1c01c6305f9c4398b5b5/websites/A/Aodesu/presence.ts#L6-L14

websites/D/Discord SFX/presence.ts
```diff
@@ -70,6 +73,22 @@ presence.on("UpdateData", async () => {
...
+		try {
+			const { searchParams } = new URL(
+				document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
+					?.src || ""
+			);
+			presenceData.largeImageKey = searchParams.get("url");
+		} catch {
+			presenceData.largeImageKey = Assets.Logo;
+		}
+	}
```
DarkVillager, line 90:
```suggestion
			const { searchParams } = new URL(
				document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
					?.src || ""
			);
			presenceData.largeImageKey = searchParams?.get("url") ?? Assets.Logo;;
	}
```

DarkVillager, line 90:
```suggestion
			const { searchParams } = new URL(
				document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
					?.src || ""
			);
			presenceData.largeImageKey = searchParams?.get("url") ?? Assets.Logo;;
	}
```

DarkVillager (changes requested):

websites/D/Discord SFX/presence.ts
chimpdev: If you don't pass a valid url to the URL, it throws an error. Shouldn't we use try catch?

DarkVillager: > If you don't pass a valid url to the URL, it throws an error. Shouldn't we use try catch?

My edited suggestion, should work.

chimpdev: Lint failed

websites/D/Discord SFX/presence.ts
```diff
@@ -70,6 +73,19 @@ presence.on("UpdateData", async () => {
...
+	if (pathname.startsWith("/users/")) {
+		presenceData.details = `Viewing ${
+			document.title.split(" Profile")[0].split(" | ")[1]
+		} profile`;
+
+			const imgEl = document.querySelector<HTMLImageElement>('[data-premid-id="avatar"]')
+					?.src,
+			{ searchParams } = imgEl ? new URL(
+				imgEl 
+			) : "";
```
theusaf, line 85:
```suggestion
			searchParams = (imgEl ? new URL(
				imgEl 
			) : null)?.searchParams;
```

Bas950 (approved):

theusaf (approved):

PR #7983, opened 2024-01-25
feat(AniMi Club): added presence

riktikdev:
## Acknowledgements

## Screenshots
<details>
https://i.imgur.com/HnGyqR6
https://i.imgur.com/54QKVEA

</details>

### DIFF
--- a/websites/A/AniMi Club/metadata.json
+++ b/websites/A/AniMi Club/metadata.json
@@ -1,0 +1,47 @@
+{
+	"$schema": "https://schemas.premid.app/metadata/1.9",
+	"author": {
+		"id": "546262814424170498",
+		"name": "official_riktik"
+	},
+	"service": "AniMi Club",
+	"description": {
+		"en": "Watch anime online for free. Large database of the best anime with russian voiceover in good quality.",
+		"ru": "Смотреть аниме онлайн бесплатно. Большая база лучших аниме с русской озвучкой в хорошем качестве.",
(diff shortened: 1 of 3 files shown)


websites/A/AniMi Club/metadata.json
```diff
@@ -0,0 +1,47 @@
...
+	"tags": [
+		"anime",
+		"аниме",
+		"animi",
+		"аними",
+		"animiclub",
+		"анимиклаб"
+	],
+	"iframe": true,
+	"iFrameRegExp": ".*",
```
theusaf, line 32: If possible, set the iframe regex to the urls that it accesses

theusaf (changes requested):
Other than this, looks good.

websites/A/AniMi Club/metadata.json
riktikdev: I'm a little confused about what you're talking about 😅

theusaf: The current value `.*` is alright, but means that it runs on any iframes on the page. We'd prefer it if the regex only matches the iframe urls that the page actually uses for its video features.

For example, if the page uses video sources from `example.com` and `foo.com`, a potential regex could be:
```json
"(example[.]com|foo[.]com)[/]"
```

riktikdev: Unfortunately there are links in the iframe that I may not be aware of. The API I use for the player may give different links.

theusaf (approved):

Bas950 (approved):

PR #8192, opened 2024-03-22
feat(Prime Video): rewrite presence

veryCrunchy:


theusaf:
Doesn't this collide with the Amazon presence? The url should already be matching through the regex

veryCrunchy:
Url isn't used if there's a regex right?
Added the url for looks.
So users can see that it actually work's when watching on amazon.com on the store page.

theusaf:
> Url isn't used if there's a regex right? Added the url for looks. So users can see that it actually work's when watching on amazon.com on the store page.

Here's the current extension's code: It uses `url` if `regExp` doesn't match.
```ts
presence = presences.find((presence) => {
				const { url, regExp } = presence;

				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
				if (regExp && new RegExp(regExp).test(tab.url!)) return true;

				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
				if (Array.isArray(url) && url.some((url) => tab.url!.includes(url)))
					return true;

				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
				if (!Array.isArray(url) && tab.url!.includes(url)) return true;

				return false;
			});
```

veryCrunchy:
I see, even then it shouldn't be an issue.
The pages the presence check for don't exist on the amazon website, so those will never match anyway

veryCrunchy:
But I should test it, thank you for bringing that up.
Will mark this pr as a draft for now.

theusaf:
Closing for inactivity. This may be reopened later if needed.

PR #8330, opened 2024-04-28
refactor(Minecraft Wiki) use mw.config

Markus-Rost:
## Description 

Reworked the Minecraft Wiki presence to use `mw.config` variables for more reliable checks.

## Acknowledgements
- [ ] I linted the code by running `yarn format`

## Screenshots
<details>

</details>


websites/M/Minecraft Wiki/presence.ts
```diff
@@ -38,27 +37,29 @@ presence.on("UpdateData", async () => {
 			btnViewThread: "apple.btnViewThread",
 			viewWatchlist: "minecraft wiki.viewWatchlist",
 		}),
+		mwConfig = await presence.getPageletiable<{
```
theusaf, line 40: Preferably, use `getPageVariable` which has a better performance and is not deprecated.

theusaf (changes requested):

websites/M/Minecraft Wiki/presence.ts
Markus-Rost: I was looking at https://docs.premid.app/en/dev/presence/class which didn't list that function being deprecated and didn't mention getPageVariable at all. How would I use that function correctly? Can I use `mw.config.get(['name','name'])` instead of `mw.config.values`?

theusaf: Yeah, the documentation is in the process of being updated. Usage looks like this:
```ts
const {
  "mw.config.values.foo": foo,
  "mw.config.values.bar": bar,
} = presence.getPageVariable([
  "mw.config.values.foo",
  "mw.config.values.bar"
]);
```

Markus-Rost: Hm, I would prefer to have the variables on a single object variable instead of them each being their own variable.

Markus-Rost: I've switched to getPageVariable

theusaf (approved):

theusaf (approved):

Issue #8131, opened 2024-02-27
YouTube Music | music.youtube.com - Correct Timestamps after resuming playback


RichardKanshen:
### Presence name

YouTube Music

### Description

After resuming playback, the timestamps get reset as if the video started midway.
![image](https://github.com/PreMiD/Presences/assets/73738591/ef6f91dd-5178-49bd-809b-58f2339503f5)
![image](https://github.com/PreMiD/Presences/assets/73738591/dab17c5c-a9e1-4408-9eb4-714434d71ffb)

### Additional context

After looking through the code of the presence, it should be fairly simple to fix, I believe just one change in the last function of the presence should fix it (snippet - YouTube Music // presence.ts - Lines 268 - 279)

```
function updateSongTimestamps() {
	const element = document
			.querySelector<HTMLSpanElement>("#left-controls > span")
			.textContent.trim()
			.split(" / "),
		[currTimes, totalTimes] = element;

	mediaTimestamps = presence.getTimestamps(
		presence.timestampFromFormat(currTimes),
		presence.timestampFromFormat(totalTimes)
	);
}
```

DarkVillager:
There always seem to be a second difference. Let me see what I can do.

DarkVillager:
Seems like it's only a 0.5 difference, that gets floored to a 1 sec diff

DarkVillager:
Doesn't seem fixable.

RichardKanshen:
The issue seems to be rather that the time resets after pausing. The song is shown shorter after pausing and resuming.

DarkVillager:
I can't reproduce that.

RichardKanshen:
I recorded it, maybe this might help?

https://github.com/PreMiD/Presences/assets/73738591/8310af05-b789-449a-8b77-4dcaf1d017a7

PR #8324, opened 2024-04-28
fix(Youtube Music): timestamps

ION606:
## Description 

* Changed all `ActivityType.Listening` to `ActivityType.Playing` so the timestamps are actually displayed
* Added the "Use Time Left" setting that allows a user to switch between displaying the time left and displaying the time elapsed

## Acknowledgements
- [x] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
- [x] I linted the code by running `yarn format`
- [x] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>

image

image

image

<summary> Proof showing the creation/modification is working as expected </summary>

</details>

Resolves https://github.com/PreMiD/Presences/issues/8262

PR #8369, opened 2024-05-11
fix(Youtube Music): timestamp fix

ION606:
## Description 

* Fixed the issue where the start timestamp would malfunction when changing from `use time left` mode.

This issue was caused due to the [getTimestamps](https://github.com/PreMiD/Extension/blob/release/src/util/presence/devHelper.ts#L527) function, which for SOME REASON completely ignores the first parameter in favor of using `Date.now()`

possibly resolves https://github.com/PreMiD/Presences/issues/8131

## Acknowledgements

## Screenshots
<details>

image

image

</details>


veryCrunchy:
Fixed in alpha/beta

ION606:
no, this PR deals with when the timestamp is updated (i.e. on seek)

veryCrunchy:
Please try the alpha/beta, this doesn't seem to be an actual issue

veryCrunchy:
I think you're better off removing the useTimeLeft toggle fully.
You're introducing more bugs for elapsed right now than that you are fixing.
Having the time left is the preferred behaviour anyway.

ION606:
I suppose, but also, check this pull and the corresponding issue https://github.com/PreMiD/Presences/pull/8324

websites/Y/YouTube Music/presence.ts
```diff
@@ -277,8 +283,21 @@ function updateSongTimestamps(useTimeLeft: boolean) {
...
+		// leaving this here until it gets fixed in the extension, just in case
+		// mediaTimestamps = [
+		// 	Date.now() / 1000 - presence.timestampFromFormat(currTimes),
+		// 	null,
+		// ];
+		mediaTimestamps = presence.getTimestamps(
+			presence.timestampFromFormat(currTimes),
+			null
+		);
+		delete mediaTimestamps[1];
```
veryCrunchy, line 301: All you're doing now is setting `Math.floor(Date.now() / 1000)` in a overcomplicated fashion.
The readable equivalent would be
```suggestion
		mediaTimestamps[0] = Math.floor(Date.now() / 1000);
```
However the startTimestamp is also already being set [here](https://github.com/PreMiD/Presences/blob/40d3e76322749e1bc94b20d68062ad4b85911f0b/websites/Y/YouTube%20Music/presence.ts#L170) of the code.
So you'll have to properly take the timestampFromFormat into account

ION606:
odd, I thought I removed that part, my fix was the toggle and the additional call on seek

image

veryCrunchy:
If you want to have the actual elapsed time of the song you'll have to calculate the songs timestamps.
The Date.now() isn't doing much now.
Also please remove and don't set the event listeners if privacy mode is enabled.

ION606:
> If you want to have the actual elapsed time of the song you'll have to calculate the songs timestamps. The Date.now() isn't doing much now. Also please remove and don't set the event listeners if privacy mode is enabled. And move all updateSongTimestamps calls behind the privacy mode statement.

wait where do I set event listeners?

websites/Y/YouTube Music/presence.ts
```diff
@@ -277,8 +283,10 @@ function updateSongTimestamps(useTimeLeft: boolean) {
...
-	mediaTimestamps = presence.getTimestamps(
-		useTimeLeft ? presence.timestampFromFormat(currTimes) : Date.now(),
-		useTimeLeft ? presence.timestampFromFormat(totalTimes) : null
-	);
+	if (useTimeLeft) {
+		mediaTimestamps = presence.getTimestamps(
+			presence.timestampFromFormat(currTimes),
+			presence.timestampFromFormat(totalTimes)
+		);
+	} else mediaTimestamps[0] = Math.floor(Date.now() / 1000);
```
veryCrunchy, line 291: Start timestamp will be "reset" each time `updateSongTimestamps` is called.
I think we want to show the elapsed time from the playing song instead.

veryCrunchy (changes requested):

websites/Y/YouTube Music/presence.ts
ION606: I'm pretty sure setting the timestamp to now does nothing (it's just a theory bc it works fine for me)

veryCrunchy: Setting the timestamp to `Date.now()` will reset the elapsed time. There's no time difference between now and now.
So each time you call it, you will reset it. which happens when a user unpauses the song or seeks.
If the user is listening to a song at 3.34 / 4.10 we want to show 3m34s elapsed, equivalent to showing 36s left.

ION606: yer I know but it....does that. I'm confused now. Can you clone it and see if it works for you or just some strange bug for me?

veryCrunchy: Right now you are setting Date.now to the start timestamp each time `updateSongTimestamps` is called.
I don't need to run the presence locally to see that it doesn't show the actual elapsed time of the song.

ION606: but it....it does. Like it works perfectly fine for me

veryCrunchy: Don't "it works on my machine" me now 😭

ION606: HDJEJDIEJDIR IT DOES

veryCrunchy: BUT ITS NOT THE BEHAVIOR WE WANT GOOD SIR

ION606: OKOK IN A BIT I WILL GO OVER IT FJENIDNEIDBEIDE

veryCrunchy: So take date.now and subtract the elapsed song time from it

websites/Y/YouTube Music/presence.ts
```diff
@@ -275,10 +281,16 @@ function updateSongTimestamps(useTimeLeft: boolean) {
...
-		useTimeLeft ? presence.timestampFromFormat(currTimes) : Date.now(),
-		useTimeLeft ? presence.timestampFromFormat(totalTimes) : null
-	);
+		[currTimes, totalTimes] = element,
+		TSNew = presence.getTimestamps(
+			presence.timestampFromFormat(currTimes),
+			presence.timestampFromFormat(totalTimes)
+		);
+	if (useTimeLeft) mediaTimestamps = TSNew;
+	else {
```
veryCrunchy, line 290:
```suggestion
	if (useTimeLeft) mediaTimestamps = presence.getTimestamps(
			presence.timestampFromFormat(currTimes),
			presence.timestampFromFormat(totalTimes)
		);
	else {
```
What's the reason for `TSNew`? 
It seems redundant, plus redundant `timestampFromFormat` calls when !useTimeLeft

veryCrunchy (changes requested):

veryCrunchy:
Is `useTimeLeftChanged` really necessary?

ION606:
> Is `useTimeLeftChanged` really necessary?

yes

veryCrunchy:
Looks good then
@theusaf (I can't request reviews ;-;)

Bas950 (approved):

theusaf (approved):

PR #8756, opened 2024-09-24
feat(Rádio Animu): add listening type

bossatossa:
## Description 
It updates the instance to use the listening activity type.

## Acknowledgements

## Screenshots
<details>

Screenshot_20240924_203731

Screenshot_20240924_213502

</details>


websites/R/Rádio Animu/presence.ts
```diff
@@ -29,6 +29,8 @@ presence.on("UpdateData", async () => {
 	};
 
 	if (playing) {
+		presenceData.type = ActivityType.Listening;
+		presenceData.name = "Rádio Animu";
```
darkvillager2, line 33: not really necessary?

darkvillager2 (changes requested):

websites/R/Rádio Animu/presence.ts
bossatossa: True, but it is a radio station, it makes sense to display it as a listening activity, you don't _play_ a station, you **listen** to it.

The name is there to avoid the situation in this screenshot, where the name isnt displayed in the popup.

Screenshot_20240925_222121

Bas950: Could you try this without the .name in the latest alpha version.
If you don't have access to it @ me in discord.

The .name should be defaulted to the presence name by the extension, so then the code is not needed.

bossatossa: Makes sense, i will try it out on the alpha version when get the access to it.

Bas950 (approved):

theusaf (approved):

PR #8779, opened 2024-10-02
feat(Docchi): regexp update and visual refresh

Meff1u:
## Description 
Added some URLs to iframe regexp and and moved details on presence to make it look better

## Acknowledgements

## Screenshots
- 
With privacy turned on
-


Meff1u:
i have no clue why there are commits from february

websites/D/Docchi/presence.ts
```diff
@@ -139,6 +140,7 @@ presence.on("UpdateData", async () => {
 							},
 						];
 					}
+					presenceData.state = "Docchi.pl";
```
darkvillager2, line 143: Not really sure if this is advertising, if it is it isn't allowed.

darkvillager2 (changes requested):
Other than this LGTM

websites/D/Docchi/presence.ts
Meff1u: It’s not meant to look like an advertisement. I set it up due to changing the presence name to 'watching series,' which means there's no longer 'Watching Docchi' while watching, and nothing besides this 'Docchi.pl' state indicates the site you're watching on. However, if this is not allowed, I’ll remove it :)

Bas950 (approved):

websites/D/Docchi/presence.ts
Bas950: It's klnda meh to me...

Meff1u: fine, then im gonna remove it

theusaf (approved):

PR #8842, opened 2024-10-26
fix(YTPLR): fix iframe has already been declared

PeterStrick:
## Description 
This PR fixes an Error only present on PreMiD 2.6.12 (and likely older Versions?) that is not present on either Alpha or Beta Versions of PreMiD 2.7.x

Said Error being `Uncaught SyntaxError: Identifier 'iframe' has already been declared (at iframe.js:1:1)`

## Acknowledgements

## Screenshots
<details>
<summary> Error Screenshot </summary>

image

</details>

### DIFF
--- a/websites/Y/YTPLR/iframe.ts
+++ b/websites/Y/YTPLR/iframe.ts
@@ -1,10 +1,10 @@
-const iframe = new iFrame();
-iframe.on("UpdateData", async () => {
+const ytplr_iframe = new iFrame();
+ytplr_iframe.on("UpdateData", async () => {
 	const videoElement: HTMLVideoElement =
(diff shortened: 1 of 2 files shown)


darkvillager2:
@PeterStrick that issue can be ignored and doesnt need to be solved.

PeterStrick:
> @PeterStrick that issue can be ignored and doesnt need to be solved. Iirc

Problem being though as far as I could test, the Presence does not work on Stable PreMiD if the variable isn't renamed which is weird

Bas950 (approved):

Timeraa (approved):
It's a weird issue with the stable version that only happens on certain cases so this is fine

PR #9065, opened 2025-01-02
fix(AnimeLib): add hack to resolve Deepscan issue in lib.ts

Bas950:
Add a no-op function to bypass Deepscan warnings. This change 
ensures that the code passes static analysis while maintaining 
functionality.


Bas950:
https://github.com/PreMiD/Presences/pull/9056

PR #9094, opened 2025-01-05
feat(Granblue Fantasy): add raid details & query only necessary data

kyrie25:
## Description 

- Improve query performance by only getting necessary data
- Fix getting main character by getting list of all characters instead of only on-field
- Add option to show profile (details), raid element, event title, turn counter

## Acknowledgements

## Screenshots
<details>

Discord_y3ZquWzBuy
Discord_lMHRgobUQX
vivaldi_Hd1zF358It

</details>

### DIFF
--- a/websites/G/Granblue Fantasy/presence.ts
+++ b/websites/G/Granblue Fantasy/presence.ts
@@ -1,4 +1,7 @@
 /* eslint-disable camelcase */
+// Hack to resolve Deepscan
+const no_op = (a: number) => a + 1;
+no_op(0);
 const presence = new Presence({
 		clientId: "632983924414349333",
 	}),
--- a/websites/G/Granblue Fantasy/presence.ts
+++ b/websites/G/Granblue Fantasy/presence.ts
@@ -1,9 +1,7 @@
 const presence = new Presence({
 		clientId: "632983924414349333",
 	}),
-	browsingTimestamp = Math.floor(Date.now() / 1000),
-	script = document.createElement("script"),
-	eventId = "PreMiD_GranBlueFantasy";
+	browsingTimestamp = Math.floor(Date.now() / 1000);
 
 enum Elements {
 	Plain,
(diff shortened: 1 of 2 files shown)


websites/G/Granblue Fantasy/presence.ts
```diff
@@ -5,65 +5,151 @@ const presence = new Presence({
...
+		[Elements.Water]:
+			"https://gbf.wiki/images/thumb/8/80/Icon_Element_Water.png/20px-Icon_Element_Water.png",
+		[Elements.Earth]:
+			"https://gbf.wiki/images/thumb/9/94/Icon_Element_Earth.png/36px-Icon_Element_Earth.png",
+		[Elements.Wind]:
+			"https://gbf.wiki/images/thumb/6/62/Icon_Element_Wind.png/20px-Icon_Element_Wind.png",
+		[Elements.Light]:
+			"https://gbf.wiki/images/thumb/1/19/Icon_Element_Light.png/20px-Icon_Element_Light.png",
+		[Elements.Dark]:
+			"https://gbf.wiki/images/thumb/a/a5/Icon_Element_Dark.png/20px-Icon_Element_Dark.png",
```
Bas950, line 33: Images should be 512x512

websites/G/Granblue Fantasy/presence.ts
```diff
@@ -5,65 +5,151 @@ const presence = new Presence({
...
+				detail: {
+					gGameStatus: null,
+					userData: null,
+					error: error.message,
+					stack: error.stack
+				}
+			}));
+			isRunning = false;
+		}
+  	}, 3000);
```
Bas950, line 122: Try using `getPageVariable`

Bas950 (changes requested):

websites/G/Granblue Fantasy/presence.ts
kyrie25: There aren't any larger images available, these are taken directly from game assets

websites/G/Granblue Fantasy/presence.ts
kyrie25: All these variables have circular dependencies, which makes the browser absolutely shits itself when `JSON.stringify` is called on it, which is also what the extension is using to send variable data.

This is why the script was used in the first place

kyrie25: References to the reasons behind the script:
https://discord.com/channels/493130730549805057/607524579874832446/930332649795493938
https://discord.com/channels/493130730549805057/607524579874832446/930390425481662504

Bas950: @Timeraa can confirm, but I am pretty sure the circular will be replaced with `"[Circular]"`

kyrie25: > @Timeraa can confirm, but I am pretty sure the circular will be replaced with `"[Circular]"`

Unless you have manually implemented the circular replacer from here, which matches the string replacement you mentioned, then yes
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value

From what I'm seeing in the current stable version of the extension (`v2.7.4`), it does not, and `JSON.stringify` does not handle these cases by default.

Regardless, I have tested it and it does *still* freezes the browser tab

kyrie25: There is also the issue of performance (as the stringified object is 20kb in size), which is why the script had to be run on a 3-4s delay `setInterval`.
This PR aims to reduce that by selecting only the few properties needed in the object, of which I don't think `getPageVariable` is capable of atm.

Bas950: I mean you can use dot props in the getPageVariables, so you only need to fetch what you need, and you can fetch multiple at once.

Bas950: For example in youtube: https://github.com/PreMiD/Presences/blob/9f824827e086587b3db9b512ff14d7c29a580436/websites/Y/YouTube/video_sources/api.ts#L28

kyrie25: Gave it a shot, it would still run into some kind of issues that blocks the extension from updating, most likely XSS/circular dependency as I've encountered using the custom script

image

websites/G/Granblue Fantasy/presence.ts
theusaf: You can use a tool like waifu2x to upscale the image

kyrie25: I am aware that you can upscale the images but that doesn't make the original assets look any better, especially for images already 25 times smaller in comparison

If you want 512x512 images for the *sake* of having 512x512 images then I can make them larger that way, but I don't see the point since they are of the same quality

websites/G/Granblue Fantasy/presence.ts
kyrie25: I somehow made that work, doesn't seem to have that much overhead either

theusaf:
The deepscan issue is a weird one. You'll have to add the following hack to resolve the issue:
```ts
/* eslint-disable camelcase */
// Hack to resolve Deepscan
const no_op = (a: number) => a + 1;
no_op(0);
```

kyrie25:
> The deepscan issue is a weird one. You'll have to add the following hack to resolve the issue:
> 
> ```ts
> /* eslint-disable camelcase */
> // Hack to resolve Deepscan
> const no_op = (a: number) => a + 1;
> no_op(0);
> ```

I'm assuming the `eslint-disable` directive is enough? Sorry I'm not familiar with how DeepScan works.

theusaf (approved):

websites/G/Granblue Fantasy/presence.ts
```diff
@@ -1,69 +1,135 @@
+/* eslint-disable camelcase */
```
Bas950, line 1:
```suggestion
/* eslint-disable camelcase */
// Hack to resolve Deepscan
const no_op = (a: number) => a + 1;
no_op(0);
```

Bas950 (changes requested):

Bas950 (approved):

websites/G/Granblue Fantasy/presence.ts
Timeraa: We do replace circular references with [Circular], this only applies to the `getPageVariable` method though.

PR #9189, opened 2025-01-21
fix(Anghami): add cover and timestamps

Aphex2in:
## Description 

Including song cover into the activity and fixing the timestamp issue.

## Acknowledgements
- [ ] I read the [Presence Guidelines](https://github.com/PreMiD/Presences/blob/main/.github/CONTRIBUTING.md)
- [ ] I linted the code by running `yarn format`
- [ ] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Presences/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>

image
image

</details>


websites/A/Anghami/presence.ts
```diff
@@ -2,43 +2,66 @@ const presence = new Presence({
 		clientId: "721740741570986016",
 	}),
 	strings = presence.getStrings({
-		play: "general.playing",
+		play: "general.play",
```
Bas950, line 5: play doesn't exist, playing does.

websites/A/Anghami/presence.ts
```diff
@@ -2,43 +2,66 @@ const presence = new Presence({
...
+			const match = decodeURIComponent(style).match(/url\("([^"]+?)"\)/);
+			if (match && match[1]) {
+				let url = match[1];
+				url = url.replace(/&size=\d+/, "");
+				return url;
+			}
+		}
+	} else {
+		const url = "https://cdn.rcd.gg/PreMiD/websites/A/Anghami/assets/logo.png";
+		return url;
```
Bas950, line 24:
```suggestion
		return "https://cdn.rcd.gg/PreMiD/websites/A/Anghami/assets/logo.png";
```

websites/A/Anghami/presence.ts
```diff
@@ -2,43 +2,66 @@ const presence = new Presence({
...
+		const style = trackCoverartDiv.getAttribute("style");
+		if (style) {
+			const match = decodeURIComponent(style).match(/url\("([^"]+?)"\)/);
+			if (match && match[1]) {
+				let url = match[1];
+				url = url.replace(/&size=\d+/, "");
+				return url;
+			}
+		}
+	} else {
```
Bas950, line 22: There is no else for the `if (style) {` and the `if (match && match[1]) {`, maybe instead of the else it should just be default if it hasn't returned yet set the logo asset?

Bas950 (changes requested):

websites/A/Anghami/presence.ts
Aphex2in: I wasn’t using the play icon because it gets in the way of the art cover (it will be just like the soundcloud presence when a song is on it just shows the art-cover) and I forgot to remove it from the strings definition. And thank you for your inputs I will fix them all later today.

theusaf (approved):

Bas950 (approved):

PR #9340, opened 2025-02-26
chore(cli): fix punycode deprecation warning

Slowlife01:
node-fetch uses older version of whatwg-url

no more warning
image

image

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -4230,31 +4230,6 @@
         }
       }
     },
-    "node_modules/node-fetch/node_modules/tr46": {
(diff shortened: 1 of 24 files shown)


Timeraa:
What's using node-fetch even? We shouldn't be needing it since fetch is built in now?

Slowlife01:
> What's using node-fetch even? We shouldn't be needing it since fetch is built in now?

Seems like discord-rpc

Bas950 (approved):

theusaf (approved):

PR #9389, opened 2025-03-11
 feat(CrazyGames): add presence for new site

SethFalco:
## Description

Adds support for CrazyGames.com and all locale variants of it.

This has 2 types of statuses:

* Browsing… - When on the site but not playing a game.
* Playing - When actually playing a game.

### Chores

I also did two chores on the side:

* Fixed typos in CONTRIBUTING.md
* This project uses npm, not Yarn. Amends the pull request template to lint with npm!

## Acknowledgements

## Screenshots
<details>

https://github.com/user-attachments/assets/f5e3bdbe-921e-4540-959a-2636cecbfda0

</details>

## Notes for Maintainers

* All of the descriptions come from the `meta[description]` tag of the site in each language.
* The `fetch` call that's done is to check if a mobile optimized (square) version of the games thumbnail is available, if so we prefer that over the desktop (rectangle) varient.
* Rather than handle the `pathname` from `document.location`, we get the English name of the resource, then parse that URL. This keeps the logic consistent across the site in all languages.
* The links to `cloud.falco.fun` are temporary! We can ofc replace them with either Imgur or CDN links before this is merged.

### UI/UX

I tried to make the presence data as similar as possible to other activities in the store. However, I think it would be better if the data was presented like the following.

This brings the focus more on the game rather than the platform, but still attributes the platform in the details section. What do you think?

### Languages

While working on this, I encountered the following build failure:

>  Language nb is not a valid language 

It is a valid language!

> Norwegian Bokmål
> 
> — https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes#nb

I've opted to remove this language key as I see pmd is not using a localization library, but rather comparing to the list of locales from your API. Would you have any qualms with rewording the error to something like:

> Language nb is not a supported language

Which makes clear it's not stating the language code doesn't exist, but rather PreMiD doesn't support this language, so the key is unnecessary?

### CDN

Currently, your contribution guide suggests using Imgur. However, I think it would make more sense to use Discord itself as the CDN so we don't have to upload assets across more third-party platforms.

Under **Rich Presence Assets**, one can upload an asset and we can use that CDN URL instead. Ideally, this asset will remain for as long as the app and Discord itself is around.

> 
> 
> — https://cdn.discordapp.com/app-assets/1348290237381611601/1351101516433588244.png

May I propose that you allow CDN URLs from Discord?

If this is allowed, the regex in the JSON Schema for asset URLs will have to allow query parameters.

## TODO

Before merging, the `cloud.falco.fun` links should be replaced with acceptable URLs. I've just left them like this for now while everything else gets reviewed. Was just hoping for an answer to my questions first!

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -154,7 +154,7 @@ A list of fields and their rules are listed below:
 
 ### **`description`**
 
-- **All** activities are **required** to have an English description regardless of the website's prefered language.
+- **All** activities are **required** to have an English description regardless of the website's preferred language.
 - Do **not** try and translate the description yourself unless you know that language, translators will modify your `metadata.json` and change the descriptions if necessary.
 
 ### **`url`**
@@ -225,7 +225,7 @@ Here is a list of rules you must follow when writing your `presence.ts` file:
 - **Always** use `document.location` to get current location information rather than `window.location` or `location`.
 - All assets **must** have a resolution of `512x512` pixels. You can upsize it using a tool like [waifu2x](http://waifu2x.udp.jp/).
 - **Never** use custom functions when [native variants are available](https://docs.premid.app/dev/presence#files-explained); this makes sure fixes on the extension level also apply to your activities. You're free to use whatever you need if you do not find them listed in the docs.
-- It is **forbidden** to code activities for a site without adding support to its primary language (for e.g., a YouTube activity coded with support only for Portueguese and Japanese, but not English itself.)
+- It is **forbidden** to code activities for a site without adding support to its primary language (for e.g., a YouTube activity coded with support only for Portuguese and Japanese, but not English itself.)
 - The `smallImageKey` and `smallImageText` fields are intended to provide additional/secondary context (such as `playing/paused` for video sites, `browsing` for regular sites, and other cases) not to promote Discord profiles or anything unrelated to PreMiD.
 - When accessing cookies for stored data, please prefix the key with `PMD_`.
 - You may only make HTTP/HTTPS requests to `premid.app` or the activity website API. If you are using external domains, you will be required to explain why it is necessary. The only allowed API to make requests is the [`Fetch API`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
--- a/websites/C/CrazyGames/presence.ts
+++ b/websites/C/CrazyGames/presence.ts
@@ -1,0 +1,61 @@
+import { ActivityType } from 'premid'
+
+const presence = new Presence({
+  clientId: '1348290237381611601',
+})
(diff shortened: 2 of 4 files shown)


SethFalco:
Regarding the build failure:

>  Language nb is not a valid language 

It is a valid language!

> Norwegian Bokmål
> 
> — https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes#nb

darkvillager2:
Why not make a different pr for the chores?

darkvillager2:
> ### CDN
> Currently, your contribution guide suggests using Imgur. However, I think it would make more sense to use Discord itself as the CDN so we don't have to upload assets across more third-party platforms.
> 
> 
> 
> Under **Rich Presence Assets**, one can upload an asset and we can use that CDN URL instead. Ideally, this asset will remain for as long as the app and Discord itself is around.
> 
> > 
> > — https://cdn.discordapp.com/app-assets/1348290237381611601/1351101516433588244.png
> 
> May I propose that you allow CDN URLs from Discord?
> 
> If this is allowed, the regex in the JSON Schema for asset URLs will have to allow query parameters.
> 

We have the premid cdn which the assets gets uploaded to so any contibutor can change the images when necessary.

websites/C/CrazyGames/presence.ts
```diff
@@ -0,0 +1,61 @@
...
+    startTimestamp: browsingTimestamp,
+    largeImageKey: ActivityAssets.Logo,
+  }
+
+  const englishUrl = document.querySelector('link[hreflang=en]')?.getAttribute('href')
+
+  if (englishUrl) {
+    const pathname = new URL(englishUrl).pathname
+
+    if (pathname.startsWith('/game/')) {
```
darkvillager2, line 31: I kinda miss what they're browsing if they aren't playing a game

darkvillager2 (changes requested):
Hello,

Thanks for your PR.

If you have any questions about the requested changes nor comments let me know.

.github/CONTRIBUTING.md
theusaf, line 1: Consider splitting the changes to the pull request template and contributing files to a separate pull request

websites/C/CrazyGames/presence.ts
```diff
@@ -0,0 +1,61 @@
+import { ActivityType } from 'premid'
+
+const presence = new Presence({
+  clientId: '1348290237381611601',
+})
+const browsingTimestamp = Math.floor(Date.now() / 1000)
+
+enum ActivityAssets {
+  Logo = 'https://cloud.falco.fun/s/G9ojdWxsnSRWyeZ/download/logo-v2.png',
```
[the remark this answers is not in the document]
theusaf: A download link should be fine

theusaf (changes requested):

PR #9522, opened 2025-04-26
feat(Anizium): Re-code and update presence

iSwesia:
## Description
Fully re-coded
Fixed all bugs

## Acknowledgements

## Screenshots
<details>

image

</details>

### DIFF
--- a/websites/A/Anizium/presence.ts
+++ b/websites/A/Anizium/presence.ts
@@ -30,10 +30,12 @@ function updatePoster() {
       seasonModeActive = true
       savedPosterUrl = animeImg ? luiUrl + animeImg : null
     }
-  } else if (pathname.includes('/title/')) {
+  }
+  else if (pathname.includes('/title/')) {
     savedPosterUrl = animeImg ? luiUrl + animeImg : null
     seasonModeActive = false
-  } else {
+  }
+  else {
     savedPosterUrl = null
     seasonModeActive = false
   }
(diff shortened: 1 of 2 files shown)


websites/A/Anizium/presence.ts
```diff
@@ -15,154 +15,167 @@ interface iframeData {
...
+
+      presence.setActivity(iframePresenceData)
     }
   }
-  presence.setActivity(presenceData)
-}
+})
+
+presence.on('UpdateData', async () => {
+  setTimeout(() => {
```
theusaf, line 88: Is there a reason that this timeout is needed?

iSwesia: Yes, to avoid poster lazy-load src. 1 Second delay is as ignoring lazy-load image src to get actual src path. 
When UpdateData is sent with no delay, the src is lazy-loaded, so the poster image is broke in the presence.

theusaf: Perhaps a cleaner way could be to wrap the `presece.on('UpdateData')` listener in the `setTimeout` with an explanatory comment, rather than the other way around?

iSwesia: I tried much ways to avoid lazy-load. But this is all i can and working perfectly.

iSwesia: > Perhaps a cleaner way could be to wrap the `presece.on('UpdateData')` listener in the `setTimeout` with an explanatory comment, rather than the other way around?

Also, i tried to set the SetTimeout just for largeimagekey part but it glitches when returning main page from watching page. So i set 1 sec delay for all paths.

theusaf (approved):

Bas950 (approved):

PR #10114, opened 2025-10-31
fix(Cineby): update URL from cineby.app to cineby.gd

M00seJayraffe:
## Acknowledgements
- [x] I read the [Activity Guidelines](https://github.com/PreMiD/Activities/blob/main/.github/CONTRIBUTING.md)
- [x] I linted the code by running `npm run lint`
- [x] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Activities/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>
<summary> Proof showing the creation/modification is working as expected </summary>

</details>

PR #10132, opened 2025-11-09
fix(Cineby): update URL

MWTGaming:
yesyesyes fix the dam url for once plsplspls

### DIFF
--- a/websites/C/Cineby/metadata.json
+++ b/websites/C/Cineby/metadata.json
@@ -5,11 +5,11 @@
     "id": "374905512661221377",
     "name": "slowlife."
   },
-  "service": "Cineby",
+  "service": "Cineby (Fixed url)",
   "description": {
     "en": "Watch Free TV Shows Online, Watch Free Movies Online"
   },
-  "url": "www.cineby.app",
+  "url": "https://www.cineby.gd/",
   "version": "1.0.4",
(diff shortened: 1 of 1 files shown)


websites/C/Cineby/metadata.json
```diff
@@ -5,11 +5,11 @@
     "id": "374905512661221377",
     "name": "slowlife."
   },
-  "service": "Cineby",
+  "service": "Cineby (Fixed url)",
```
NexonLogistics, line 8:
```suggestion
  "service": "Cineby",
```

websites/C/Cineby/metadata.json
```diff
@@ -5,11 +5,11 @@
...
   "description": {
     "en": "Watch Free TV Shows Online, Watch Free Movies Online"
   },
-  "url": "www.cineby.app",
+  "url": "https://www.cineby.gd/",
```
NexonLogistics, line 12:
```suggestion
  "url": "www.cineby.gd",
```

skullysmods (changes requested):
Same as #10114 

> Even after this change, the activity will not work because the Activity retrieves data using an API (which is now obsolete).
The Activity needs to be completely reworked.
> https://github.com/PreMiD/Activities/blob/d214fd768aa686cc0d88552efbcebf644d632662/websites/C/Cineby/api.ts#L38-L42

Slowlife01:
#10137

PR #10303, opened 2025-12-28
feat(Primeshows): add activity

friday2su:
## Description
Added a new Presence for [primeshows.live](https://www.primeshows.live), a Movies & Tv shows streaming service.
The presence supports:

- Detecting the show/movie title, episode information (season and episode numbers), and channel name.
- Displaying the channel logo as the large image (with fallback to service logo).
- Support for both TV series (with season/episode details) and movies.

### DIFF
--- a/websites/0-9/4Anime/iframe.ts
+++ b/websites/0-9/4Anime/iframe.ts
@@ -1,0 +1,14 @@
+const iframe = new iFrame()
+
+iframe.on('UpdateData', () => {
+  const video = document.querySelector<HTMLVideoElement>('video')
+  if (!video || Number.isNaN(video.duration))
+    return
+  iframe.send({
+    iFrameVideoData: {
+      currTime: video.currentTime,
+      dur: video.duration,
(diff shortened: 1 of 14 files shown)


websites/P/Primeshows/presence.ts
```diff
@@ -0,0 +1,327 @@
...
+      presenceData.state = `Looking for: ${query} ✨`
+    }
+    presenceData.smallImageKey = Assets.Search
+  }
+
+  if (presenceData.details !== 'Unsupported Page') {
+    presence.setActivity(presenceData)
+  }
+  else {
+    presence.setActivity()
```
theusaf: Use `clearActivity` instead to hide the Activity.

theusaf (changes requested):

websites/P/Primeshows/presence.ts
Bas950: Agreed

Bas950 (approved):

PR #10423, opened 2026-01-20
perf(Apple Music): simplify page variable dependency

### LINKED ISSUE #10419: Apple Music | music.apple.com - activity not functioning correctly
### Steps to Reproduce

Install the Apple Music activity
go to https://music.apple.com/us/home
Start playing some music

### Expected Behavior

The activity shows up on Discord

### Actual Behavior

It shows that it picked up the activity but never shows on discord

### Page URL

https://music.apple.com/us/home

### Country

United States

### Browser and Version

Chrome version 144.0.7559.60 (Official Build) (64-bit)

### Screenshots

_No response_

### Additional Comments

_No response_

kyrie25:
## Description

Resolve #10419 
`getPageVariable` is borderline unusable on Chrome as it is **EXTREMELY SLOW**, even for primitive values (e.g. `document.title`)
Presence works normally on Firefox otherwise

## Acknowledgements

## Screenshots
<details>

image

</details>

### DIFF
--- a/websites/A/Apple Music/metadata.json
+++ b/websites/A/Apple Music/metadata.json
@@ -82,6 +82,12 @@
       "value": true
     },
     {
+      "id": "buttons",
+      "title": "Show buttons",
+      "icon": "fas fa-compress-arrows-alt",
+      "value": true
+    },
+    {
       "id": "listening",
       "title": "Enable listening",
       "icon": "fad fa-ear",


theusaf (dismissed):

kyrie25:
> The slowness from the getPageVariable seems to stem from the attempt to fetch the whole `audioPlayer` object. A possible way to resolve this is to only fetch specific attributes which the Activity uses, like `audioPlayer._paused`, etc.

> it is EXTREMELY SLOW, even for primitive values (e.g. `document.title`)

Meaning even when I call this method on variables with primitive values such as `document.title` it would still time out updates.

image

When trying to fetch page variables outside the `UpdateData` cycle, the method would throw an error.
(The snippet below serves as a simple demonstration, not my actual implementation)

image

It's not a matter of trying to optimize when doing so really makes no difference. The method is unusable.

> Also, don't forget to attach screenshots showing the changes working.

Screenshot added

kyrie25:
Expanding on the topic, I can only replicate the issue on Chrome, while Firefox runs normally, which was where I tested the presence for my last PR

image

Commits in the future patches of the extension might have broke it? Regardless the current version of the presence is not usable for Chrome users atm

theusaf:
Confirmed it is an issue with the Chromium build of the extension. I've created a ticket on the support channel

Timeraa:
We found the issue in the extension regarding this, apparently `esbuild` changed part of their implementation on which we relied so we had to change it, we will release that update very soon, just checking if we can fix some other minor issues as well but getPageVariable should be very performant when it doesn't have to JSON.stringify as I do use complex logic in there to only do the slow JSON methods when it cannot copy the object so best is to get the exact value directly as @theusaf suggested. Maybe we should hold on merging this to see with the extension update (v2.12.3) @Bas950 ?

Bas950:
Converted to a draft, lets see in 2.12.3 if it fixes it for chromium.

Bas950 (approved):

Timeraa (approved):

PR #10450, opened 2026-01-24
fix(niconico): ad detection & remove unused iframe

SUKEsann2000:
## Description

In this Pull Request, I have made the following changes to the niconico activity:

- Fixed the ad detection logic, which was previously broken. Ads are now detected using the characteristic behavior of niconico videos: the controller cannot be operated during ads, instead of relying on the ad iframe.
- Removed the unused iframe-related processing (`iframe` and `iFrameRegExp`) from the metadata.json file.

## Acknowledgements

- [✅] I read the [Activity Guidelines](https://github.com/PreMiD/Activities/blob/main/.github/CONTRIBUTING.md)
- [✅] I linted the code by running `npm run lint`
- [✅] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Activities/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots

<details>
image
image
</details>


websites/N/niconico/metadata.json
theusaf, line 1: Update Activity version

websites/N/niconico/presence.ts
```diff
@@ -1,185 +1,151 @@
...
-
-  const [privacy, buttons, ads, pause, playback] = await Promise.all([
-    presence.getSetting<boolean>('privacy'),
-    presence.getSetting<boolean>('buttons'),
-    presence.getSetting<boolean>('ads'),
-    presence.getSetting<boolean>('pause'),
-    presence.getSetting<boolean>('playback'),
-  ])
+  // initialize only once
+  const initialized = await initialize(presence)
```
theusaf: Wouldn't this just return `undefined` every time after the first event? This would also prevent the strings from being updated when the user changes their extension language. You can safely call `getStrings` in each event loop as the strings are loaded and cached by the extension.

websites/N/niconico/metadata.json
SUKEsann2000: ok updated in 1f6d673

websites/N/niconico/presence.ts
SUKEsann2000: fixed in 549b044

Bas950 (approved):

PR #10522, opened 2026-02-28
feat: add allowURLOverrides to self-hostable services

Timeraa:
## Summary

- Added `"allowURLOverrides": true` to 6 self-hostable service activities: Jellyfin, Emby, Navidrome, Gitea, GitLab, and Kodi
- Bumped patch version for each updated service to signal the change
- Plex already had this field and was not modified

This allows users running self-hosted instances to configure a custom URL so the presence matches their own server rather than the default cloud URL.

## Services Updated

| Service | Version |
|---------|---------|
| Jellyfin | 1.10.0 → 1.10.1 |
| Emby | 1.7.0 → 1.7.1 |
| Navidrome | 1.2.0 → 1.2.1 |
| Gitea | 1.3.0 → 1.3.1 |
| GitLab | 1.2.0 → 1.2.1 |
| Kodi | 1.1.1 → 1.1.2 |

## Test plan

- [ ] Verify `grep -r "allowURLOverrides" websites/` returns all 7 services (Plex + 6 updated)
- [ ] Confirm each metadata.json has valid JSON (no trailing commas)
- [ ] Confirm version bumps are correct patch increments

### DIFF
--- a/websites/P/Plex/metadata.json
+++ b/websites/P/Plex/metadata.json
@@ -13,7 +13,7 @@
   },
   "url": "www.plex.tv",
   "regExp": "^https?[:][/][/](([a-z0-9-]+[.])*plex[.]tv[/])|( *[/]web[/]index[.]html#)",
-  "version": "1.11.0",
+  "version": "1.11.1",
   "logo": "https://cdn.rcd.gg/PreMiD/websites/P/Plex/assets/logo.png",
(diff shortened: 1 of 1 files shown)


websites/P/Plex/metadata.json
Bas950:
```suggestion
  "version": "1.11.1",
```

Bas950 (approved):

PR #10527, opened 2026-03-01
fix(Hulu): Watching status, menus showing incorrect/no labels.

moyaimoment:
## Description
Hulu had a lot of issues with it's Rich Presence
Here's my list of changes:
- Fixed the /watch so when your watching a show it actually showed the show length, done a weird way but works
-- I attempted to use the built in PreMiD "getTimestampFromMedia" function but it ended up not working, so instead go the current timestamp from the time under the scrubber and the Attribute from the entire slider which was conveniently in seconds.
-- This means that the Live always showed it ended at 30 minutes, so to go around that it checks for the "Go to Live" button in the video bar and just checks for the word "Live" in case the user has skipped back in the Live.

- The tabs for /show /movie and a few others tried to use the Big logo as the text to get, but it's instead in the .alt, while I could use that not every show and movie has an image for it, but at the top of every Show/Movie is the title of it in the bar as "SimpleModalNav__title" which goes for almost every popup page.

- Fixed the "Search" string to one that is actually in PreMiD still.

## Acknowledgements

## Screenshots
<details>
image
image
image
image
</details>

### DIFF
--- a/websites/H/Hulu/metadata.json
+++ b/websites/H/Hulu/metadata.json
@@ -35,7 +35,7 @@
     "www.hulu.com"
   ],
   "regExp": "^https?[:][/][/](www[.])?hulu[.]com[/]",
-  "version": "1.4.0",
+  "version": "1.5.0",
   "logo": "https://cdn.rcd.gg/PreMiD/websites/H/Hulu/assets/logo.png",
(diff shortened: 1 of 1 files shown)


moyaimoment:
Right, forgot about that

theusaf (approved):

PR #10629, opened 2026-03-31
feat(Twitch): change name to be Twitch instead of the streamers channel name

S1l3ntStr1ke87:
Changed name to be Twitch instead of the streamers channel name

## Description
Changed name to be Twitch instead of the streamers channel name, for use with stuff like https://github.com/S1l3ntStr1ke87/lanyard-typescript-template

## Acknowledgements
- [X] I read the [Activity Guidelines](https://github.com/PreMiD/Activities/blob/main/.github/CONTRIBUTING.md)
- [X] I linted the code by running `npm run lint`
- [X] The PR title follows the repo's [commit conventions](https://github.com/PreMiD/Activities/blob/main/.github/COMMIT_CONVENTION.md)

## Screenshots
<details>
<summary> Proof of it working on discord & website using lanyard </summary>
image
image
</details>


joerkig:
First, there is a setting to disable this already so it won't have the streamer as title.

Second the real way to fix this would be to update the activity to use StatusDisplayType so that it can keep it showing the streamer but won't affect the actual name of the activity and thus won't interfere with stuff like lanyard

PR #10933, opened 2026-06-24
feat(types): execInPage, onRequest & party APIs + supports() capability helper

Timeraa:
## What

Adds the activity-facing types and tooling for three new `Presence`/`iFrame` capabilities, plus a runtime feature-detection helper so activities can adopt them safely today.

- **`execInPage`** — run a function (or a declarative, CSP-immune `get`/`call` spec) in the page's own context and await a returned value.
- **`onRequest`** — read data out of page network requests via a URL/method filter (read-only, no modification).
- **`party`** on `PresenceData` — show a party size on Playing activities (e.g. "2 of 5"); `partyId` optional, only honored on `ActivityType.Playing`.
- **`supports(target, feature)`** — bundled into the activity, returns `true` when the running extension exposes the capability. Lets activities feature-detect instead of crashing on older installs.
- **ESLint rule** — errors on unguarded `execInPage`/`onRequest` usage, with an autofix that wraps the call in a `supports()` check.

## Versioning

These APIs ship with the extension in **2.14.0**. They can be **used as of right now**: because `supports()` is compiled into the activity bundle, it detects the capability at runtime, so an activity guarded with `supports()` works on 2.14.0+ and degrades gracefully (no-op) on older extensions instead of throwing. The lint rule enforces that guard.

```ts
if (supports(presence, 'onRequest')) {
  presence.onRequest({ url: '/api/now-playing', method: 'GET' }, (req) => { /* ... */ })
}
```

### DIFF
--- a/@types/premid/index.d.ts
+++ b/@types/premid/index.d.ts
@@ -749,9 +749,9 @@ declare global {
      * ```
      * @since 2.14
      */
-    execInPage<T = unknown>(
-      fn: (...args: any[]) => T | Promise<T>,
-      ...args: unknown[]
+    execInPage<T = unknown, A extends unknown[] = unknown[]>(
+      fn: (...args: A) => T | Promise<T>,
+      ...args: A
     ): Promise<T>
     /**
      * Declarative, CSP-immune variant of `execInPage`. Reads a value (`get`) or
@@ -928,9 +928,9 @@ declare global {
      * extensions don't throw.
      * @since 2.14
      */
-    execInPage<T = unknown>(
-      fn: (...args: any[]) => T | Promise<T>,
-      ...args: unknown[]
+    execInPage<T = unknown, A extends unknown[] = unknown[]>(
+      fn: (...args: A) => T | Promise<T>,
+      ...args: A
     ): Promise<T>
     execInPage<T = unknown>(spec: ExecInPageSpec): Promise<T>
     /**
--- a/@types/premid/index.d.ts
+++ b/@types/premid/index.d.ts
@@ -454,9 +454,9 @@ declare global {
     call?: string
     /** Arguments passed to the `call` function (must be serializable). */
     args?: unknown[]
-    /** Only keep these top-level keys of the result. */
+    /** Keep only these keys of the result. Dot-paths allowed, e.g. `'track.name'`. */
     pick?: string[]
-    /** Drop these top-level keys from the result. */
+    /** Drop these keys from the result. Dot-paths allowed, e.g. `'track.art'`. */
     omit?: string[]
   }
 
--- a/@types/premid/index.d.ts
+++ b/@types/premid/index.d.ts
@@ -726,6 +726,14 @@ declare global {
      *   const s = window.spotifyPlayer.getCurrentState()
      *   return { id, title: s.track.name, paused: s.paused }
      * }, userId)
+     * @remarks
+     * Added in extension 2.14. Older versions lack this method, so feature-detect
+     * before calling to avoid throwing on outdated extensions:
+     * ```ts
+     * if (typeof presence.execInPage === 'function') {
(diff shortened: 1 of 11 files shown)


@types/premid/index.d.ts
```diff
@@ -662,6 +724,74 @@ declare global {
...
+     * if (supports(presence, 'execInPage')) {
+     *   const data = await presence.execInPage(() => window.appState)
+     * }
+     * ```
+     * @since 2.14
+     */
+    execInPage<T = unknown>(
+      fn: (...args: any[]) => T | Promise<T>,
+      ...args: unknown[]
+    ): Promise<T>
```
Bas950:
```suggestion
    execInPage<T = unknown, A = unknown[]>(
      fn: (...args: A) => T | Promise<T>,
      ...args: A
    ): Promise<T>
```

Can the args not be typed like this? so that way they are passed along with their types to the function?

@types/premid/index.d.ts
```diff
@@ -789,6 +919,20 @@ declare global {
...
+     * @remarks
+     * Added in extension 2.14 — feature-detect with the bundled `supports`
+     * helper (`supports(iframe, 'execInPage')`) before calling so older
+     * extensions don't throw.
+     * @since 2.14
+     */
+    execInPage<T = unknown>(
+      fn: (...args: any[]) => T | Promise<T>,
+      ...args: unknown[]
+    ): Promise<T>
```
Bas950: ditto

Bas950 (changes requested):

@types/premid/index.d.ts
Timeraa: Good call — done in 46efb028b. Typed the rest args as a tuple generic so the values flow into the closure with their types:

```ts
execInPage<T = unknown, A extends unknown[] = unknown[]>(
  fn: (...args: A) => T | Promise<T>,
  ...args: A
): Promise<T>
```

Had to add the `A extends unknown[]` constraint (vs. plain `A = unknown[]`) — a rest parameter `...args: A` requires `A` to be array-typed, otherwise TS errors. Verified inference end-to-end: `execInPage((id, name) => ..., 42, 'x')` now types `id: number`, `name: string`, and mismatched args are rejected.

@types/premid/index.d.ts
Timeraa: Done in 46efb028b — same tuple-generic fix applied to the iFrame overload.

Bas950 (approved):

skullysmods (approved):

@types/premid/index.d.ts
```diff
@@ -429,6 +441,56 @@ declare global {
...
+    /** Dot-path to a value on `window` to read, e.g. `'player.track.name'`. */
+    get?: string
+    /** Dot-path to a page function to invoke, e.g. `'spotifyPlayer.getState'`. */
+    call?: string
+    /** Arguments passed to the `call` function (must be serializable). */
+    args?: unknown[]
+    /** Only keep these top-level keys of the result. */
+    pick?: string[]
+    /** Drop these top-level keys from the result. */
+    omit?: string[]
```
Bas950: can these be Dot-paths too?

Timeraa: Good call — done. `pick`/`omit` now accept dot-paths (not just top-level keys), matching `get`/`call`.

Runtime change in PreMiD/Extension#780: `resolvePickOmit` routes keys through `dot-prop`, so e.g. `pick: ['track.name']` narrows a `call` result to `{ track: { name } }` — which previously had no CSP-immune form (`get` can't invoke, top-level `pick` keeps the whole `track`). The filter runs on the serialized clone, so dot-path `omit` can't mutate page objects. Types + docs updated here in cac217ece.

PR #11031, opened 2026-07-20
fix(404Anime): read presence data from the DOM instead of an async bridge

sankeer28:
## What was wrong
Discord's status was never updating for people watching on 404Anime. `UpdateData` awaited an async bridge read (`execInPage`/`getPageVariable`) on every tick to get the page's presence data. `UpdateData` fires far more often than that round trip can finish, so calls piled up and the extension logged `UpdateData execution is too slow! Skipped N updates`, skipping most updates.

## The fix
Read the data from a hidden DOM element the site renders instead of an async bridge call. This matches how every other activity in this repo reads its data. It's synchronous, so `UpdateData` never awaits anything and can't fall behind.

## Acknowledgements
- [x] I linted the code by running `npx eslint "websites/0-9/404Anime" --cache`

## Tested and works
image

### DIFF
--- a/websites/0-9/404Anime/metadata.json
+++ b/websites/0-9/404Anime/metadata.json
@@ -11,7 +11,7 @@
   },
   "url": "404anime.ca",
   "regExp": "^https?[:][/][/]404anime[.]ca[/]",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "logo": "https://cdn.rcd.gg/PreMiD/websites/0-9/404Anime/assets/logo.png",
   "thumbnail": "https://cdn.rcd.gg/PreMiD/websites/0-9/404Anime/assets/thumbnail.png",
   "color": "#facc15",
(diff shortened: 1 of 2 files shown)


skullysmods (changes requested):
Bump version in metadata

skullysmods (dismissed):

Bas950 (commented):
Honestly the async one is probably better but it should be cached so it doesnt keep requesting it if there are no changes.

sankeer28:
Added some caching

Bas950 (commented):
I think you understood me wrong, I think the execInPage was the better way to get the data, but when you do use that you need to either cache it, or only request it on certain dom changes, like url change or a mutationobserver

Bas950 (changes requested):
LGTM but the ActivityChanged code can be removed. that part exists in premid itself

Bas950 (approved):

skullysmods (approved):

PR #11102, opened 2026-08-11
feat(HeiwaStream): improve navigation and episode formatting

heiwafr:
## What changed

- remove the retired `zelyon.xyz` domain from the activity metadata, URL matcher, and trusted host list
- add an `Open HeiwaStream` button while browsing, linked to the sanitized current page URL exposed by the website
- make the activity details, state, and large image open the current HeiwaStream page
- display season and episode as a compact value such as `S01E01` instead of `S01 · E01`
- bump the activity version to `1.0.3`

## Why

`zelyon.xyz` is no longer served and only redirects to HeiwaStream. Browsing presences had no direct action to return to the page, and season/episode metadata was visually split despite representing one episode coordinate.

## User impact

The Rich Presence now links back to the exact public page being browsed and presents episode coordinates in the familiar compact format.

## Validation

- `npx pmd build "HeiwaStream" --validate`
- verified that no `zelyon` reference remains under `websites/H/HeiwaStream`

### DIFF
--- a/websites/H/HeiwaStream/presence.ts
+++ b/websites/H/HeiwaStream/presence.ts
@@ -60,7 +60,6 @@ function mettreAJourActivite(): void {
       presenceData.detailsUrl = pageUrl
       presenceData.stateUrl = pageUrl
       presenceData.largeImageUrl = pageUrl
-      presenceData.buttons = [{ label: 'Open HeiwaStream', url: pageUrl }]
     }
 
     presence.setActivity(presenceData)
--- a/websites/H/HeiwaStream/presence.ts
+++ b/websites/H/HeiwaStream/presence.ts
@@ -13,8 +13,6 @@ enum ActivityAssets {
 const supportedHosts = new Set([
   'heiwastream.fr',
   'www.heiwastream.fr',
-  'zelyon.xyz',
-  'www.zelyon.xyz',
 ])
 
 function lireNombre(value: string | undefined): number | null {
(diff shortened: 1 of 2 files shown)


theusaf (approved):

websites/H/HeiwaStream/presence.ts
```diff
@@ -56,14 +55,25 @@ function mettreAJourActivite(): void {
       smallImageText: 'Navigation',
       startTimestamp: lireNombre(data.premidSince) ?? browsingTimestamp,
     }
+
+    if (pageUrl) {
+      presenceData.detailsUrl = pageUrl
+      presenceData.stateUrl = pageUrl
+      presenceData.largeImageUrl = pageUrl
+      presenceData.buttons = [{ label: 'Open HeiwaStream', url: pageUrl }]
+    }
```
theusaf, line 63: Generic buttons like this can lead to the home page which is not allowed

theusaf (changes requested):

websites/H/HeiwaStream/presence.ts
heiwafr: @theusaf, your requested change has been accepted and implemented. The generic `Open HeiwaStream` browsing button has been removed. Thanks for the review!

theusaf (approved):

skullysmods (approved):

PR #11178, opened 2026-09-07
fix(Peacock): resolve playback presence detection

hashking710:
## Description

Fix Peacock playback presence detection when the player is not exposed as a page-level video element. Playback routes now fall back to the visible title and report a Watching activity instead of remaining on Browsing. Known muted preview videos are ignored when a page-level player is available.

This also includes the existing Peacock settings for privacy, browsing status, timestamps, cover art, title display, and play/pause state.

## Validation

- Metadata version updated from 2.0.0 to 2.0.1.
- Metadata JSON parses successfully.
- Focused diagnostics pass.

### DIFF
--- a/websites/P/Peacock/metadata.json
+++ b/websites/P/Peacock/metadata.json
@@ -13,6 +13,10 @@
     {
       "name": "not__bob",
       "id": "1446263635268735151"
+    },
+    {
+      "name": "oncefeltloved",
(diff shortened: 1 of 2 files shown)


skullysmods (commented):
Upload proof of it working

skullysmods (approved):

Bas950 (approved):

